PR #51, opened 2021-09-09
PB-1856 Read the values from kdmp-config configmap instead of env

prabhu-px:
PB-1856 Read the values from kdmp-config configmap instead of env

Closes # PB-1856

### DIFF
--- a/pkg/drivers/utils/utils.go
+++ b/pkg/drivers/utils/utils.go
@@ -9,11 +9,17 @@ import (
 	"github.com/portworx/kdmp/pkg/drivers"
 	"github.com/portworx/kdmp/pkg/version"
 	"github.com/portworx/sched-ops/k8s/core"
+	"github.com/sirupsen/logrus"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/validation"
+)
+
+const (
+	defaultPXNamespace = "kube-system"
+	kdmpConfig         = "kdmp-config"
 )
 
 // NamespacedName returns a name in form "<namespace>/<name>".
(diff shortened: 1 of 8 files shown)


pkg/drivers/utils/utils.go
```diff
@@ -73,9 +74,22 @@ func ToJobStatus(progress float64, errMsg string) *drivers.JobStatus {
 	}
 }
 
+// GetConfigValue read configmap and return the value of the requested parameter
+func GetConfigValue(key string) string {
+	configMap, err := core.Instance().GetConfigMap(
+		"kdmp-config",
+		"kube-system",
+	)
+	if err != nil {
```
ram-infrac, line 83: it should be - 
```
	if err != nil {
			log.Errorf("Failed to read configmap. :%v",err)
	                return ""
	}
        return configMap.Data[key]
}
```

pkg/drivers/utils/utils.go
```diff
@@ -73,9 +74,22 @@ func ToJobStatus(progress float64, errMsg string) *drivers.JobStatus {
 	}
 }
 
+// GetConfigValue read configmap and return the value of the requested parameter
+func GetConfigValue(key string) string {
+	configMap, err := core.Instance().GetConfigMap(
+		"kdmp-config",
```
ram-infrac, line 80: make this constant

ram-infrac (approved):
lgtm after fixing comments

prashanthpx (approved):
looks good to me (no other comments apart from the current ones)

pkg/drivers/utils/utils.go
```diff
@@ -73,9 +78,22 @@ func ToJobStatus(progress float64, errMsg string) *drivers.JobStatus {
...
+func GetConfigValue(key string) string {
+	configMap, err := core.Instance().GetConfigMap(
+		kdmpConfig,
+		defaultPXNamespace,
+	)
+	if err != nil {
+		return ""
+	}
+	log.Errorf("Failed to read configmap.")
+	return configMap.Data[key]
```
adityadani, line 91: If a config map is not found, can we check if there is an env variable?

pkg/drivers/utils/utils.go
siva-portworx: Fixed

pkg/drivers/utils/utils.go
siva-portworx: Fixed

pkg/drivers/utils/utils.go
siva-portworx: Fixed

prashanthpx (approved):
lgtm

PR #52, opened 2021-09-09
pb-1859: Added changes for kopia snapshot delete functionality.

siva-portworx:
 pb-1859: Added changes for kopia snapshot delete functionality.

        - Added kopia snapshot delete driver
        - Added executor support to invoke kopia snapshot delete
          command.
          
Closes # PB-1859

**Testing:**
1) Vendor the kdmp vendor code to px-backup. 
2) Added the logic to delete the snapshot, if the driver type is kdmp.
3) Created a normal backup object with few PVC.
4) Created a kopia snapshot offline with a repo.
5) Took the snapshot ID and updated the one of the volume of the backup object to have this snapshot ID and driver type as kdmp.
6) Then tried to deleted the backup object from the UI.
7) Made sure the snapshot is deleted after the completion of the backup object delete.
8) Also verified that delete and container are created and it's content.
9) jobs are deleted automatically after the completion of the snapshot delete.

### DIFF
--- a/pkg/drivers/kopiadelete/kopiadelete.go
+++ b/pkg/drivers/kopiadelete/kopiadelete.go
@@ -1,0 +1,233 @@
+package kopiadelete
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/portworx/kdmp/pkg/drivers"
+	"github.com/portworx/kdmp/pkg/drivers/utils"
+	"github.com/portworx/sched-ops/k8s/batch"
+	"github.com/sirupsen/logrus"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	kopiaDeleteJobPrefix = "snapshot-delete"
+)
+
+// Driver is a kopia delete snapshot implementation
+type Driver struct{}
+
+// Name returns a name of the driver.
+func (d Driver) Name() string {
+	return drivers.KopiaDelete
+}
+
+// StartJob creates a job for kopia snapshot delete
+func (d Driver) StartJob(opts ...drivers.JobOption) (id string, err error) {
+	fn := "StartJob:"
+	o := drivers.JobOpts{}
+	for _, opt := range opts {
+		if opt != nil {
+			if err := opt(&o); err != nil {
+				return "", err
+			}
+		}
+	}
+
+	if err := d.validate(o); err != nil {
+		errMsg := fmt.Sprintf("validation failed for snapshot delete job for snapshotID [%v]: %v", o.SnapshotID, err)
+		logrus.Infof("%s %v", fn, errMsg)
+		return "", fmt.Errorf(errMsg)
+	}
+	jobName := toJobName(o.SnapshotID)
+	job, err := buildJob(jobName, o)
+	if err != nil {
+		errMsg := fmt.Sprintf("building backup snapshot delete job [%s] failed: %v", jobName, err)
+		logrus.Errorf("%s %v", fn, errMsg)
+		return "", fmt.Errorf(errMsg)
+	}
+	if _, err = batch.Instance().CreateJob(job); err != nil && !apierrors.IsAlreadyExists(err) {
+		errMsg := fmt.Sprintf("creation of backup snapshot delete job [%s] failed: %v", jobName, err)
+		logrus.Errorf("%s %v", fn, errMsg)
+		return "", fmt.Errorf(errMsg)
+	}
+	logrus.Infof("%s created backup snapshot delete job [%s] successfully", fn, job.Name)
+	return utils.NamespacedName(job.Namespace, job.Name), nil
+}
+
+// DeleteJob delete the backup snapshot delete job.
+func (d Driver) DeleteJob(id string) error {
+	fn := "DeleteJob:"
+	namespace, name, err := utils.ParseJobID(id)
+	if err != nil {
+		logrus.Errorf("%s %v", fn, err)
+		return err
+	}
+
+	if err = batch.Instance().DeleteJob(name, namespace); err != nil && !apierrors.IsNotFound(err) {
+		errMsg := fmt.Sprintf("deletion of delete snapshot job [%s/%s] failed: %v", namespace, name, err)
+		logrus.Errorf("%s: %v", fn, errMsg)
+		return fmt.Errorf(errMsg)
+	}
+
+	return nil
+}
+
+// JobStatus returns a progress status for a data transfer.
+func (d Driver) JobStatus(id string) (*drivers.JobStatus, error) {
+	fn := "JobStatus"
+	namespace, name, err := utils.ParseJobID(id)
+	if err != nil {
+		return utils.ToJobStatus(0, err.Error()), nil
+	}
+
+	job, err := batch.Instance().GetJob(name, namespace)
+	if err != nil {
+		errMsg := fmt.Sprintf("failed to fetch backup %s/%s job: %v", namespace, name, err)
+		logrus.Errorf("%s: %v", fn, errMsg)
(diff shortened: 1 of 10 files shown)


pkg/drivers/kopiadelete/kopiadelete.go
```diff
@@ -0,0 +1,241 @@
...
+					Labels: labels,
+				},
+				Spec: corev1.PodSpec{
+					RestartPolicy:    corev1.RestartPolicyOnFailure,
+					ImagePullSecrets: utils.ToImagePullSecret(utils.KopiaExecutorImageSecret()),
+					Containers: []corev1.Container{
+						{
+							Name:            "kopiaexecutor",
+							Image:           utils.KopiaExecutorImage(),
+							ImagePullPolicy: corev1.PullAlways,
```
prashanthpx, line 175: Change to `PullIfNotPresent` we don't want to Pull the image for every job

pkg/executor/kopia/kopiadelete.go
```diff
@@ -0,0 +1,116 @@
...
+	// Check if kopia.repository exists
+	exists, err := isRepositoryExists(repo)
+	if err != nil {
+		errMsg := fmt.Sprintf("repository exists check for repo [%s] failed: %v", repo.Name, err)
+		logrus.Errorf("%s: %v", fn, errMsg)
+		return fmt.Errorf(errMsg)
+	}
+
+	if !exists {
+		if err = runKopiaCreateRepo(repo); err != nil {
```
prashanthpx, line 55: We don't need repo creation if it doesn't exist. We just need to bail out from the top.

pkg/executor/kopia/kopiadelete.go
```diff
@@ -0,0 +1,116 @@
...
+	deleteCmd, err := kopia.GetDeleteCommand(
+		snapshotID,
+	)
+	if err != nil {
+		errMsg := fmt.Sprintf("getting delete backup snapshot command for snapshot ID [%v] failed: %v", snapshotID, err)
+		logrus.Errorf("%s %v", fn, errMsg)
+		return fmt.Errorf(errMsg)
+	}
+	// This is needed to handle case where after kopia repo create was successful and
+	// the pod got terminated. Now user triggers another backup, so we need to pass
```
prashanthpx, line 88: This comment is not needed here right? This comment can also be removed from kopiabackup.go file.

pkg/kopia/command.go
```diff
@@ -156,3 +158,24 @@ func (c *Command) SetPolicyCmd() *exec.Cmd {
 
 	return cmd
 }
+
+// DeleteCmd returns os/exec.Cmd object for the kopia snapshot delete Command
+func (c *Command) DeleteCmd() *exec.Cmd {
+	// Get all the flags
+	argsSlice := []string{
+		"snapshot",
+		c.Name, // connect command
```
prashanthpx, line 167: nit: comment should be `delete` command

pkg/kopia/delete.go
```diff
@@ -0,0 +1,82 @@
...
+
+	if err := d.execCmd.Start(); err != nil {
+		d.lastError = err
+		return err
+	}
+	d.isRunning = true
+	go func() {
+		err := d.execCmd.Wait()
+		if err != nil {
+			d.isRunning = false
```
prashanthpx, line 53: nit: Can be moved after line #51 and also removed from line #59 as it's always false in either case.

pkg/drivers/kopiadelete/kopiadelete.go
```diff
@@ -0,0 +1,241 @@
...
+		errMsg := fmt.Sprintf("deletion of delete snapshot job [%s/%s] failed: %v", namespace, name, err)
+		logrus.Errorf("%s: %v", fn, errMsg)
+		return fmt.Errorf(errMsg)
+	}
+
+	return nil
+}
+
+// JobStatus returns a progress status for a data transfer.
+func (d Driver) JobStatus(id string) (*drivers.JobStatus, error) {
```
prashanthpx, line 90: Do we need `JobStatus()` as this works on VokumeBackup CR which we don't have in the Delete path. For eg the `id` is fetched from VB CR to know which job to delete.

siva-portworx: Yes, in delete case, we will not get the status from volumeBackup CR. But using JobStatus interface API to get the status of the job resource and analysis the status and return to the caller the status of success, failure or Inprogress. Otherwise, I have to read the job directly in the px-backup and do the analysis.

pkg/drivers/kopiadelete/kopiadelete.go
siva-portworx: Yes, i changes it to for debugging purpose as I want to download the changed image always. I was thinking of having it for some time. So that it is easy for debugging. Later when it gets stabilises, we can revert it to NotPresent. Will add a TODO.

pkg/kopia/command.go
siva-portworx: Fixed.

pkg/kopia/delete.go
siva-portworx: I can not move it to line number to 51 because we want to make the isRunning false only when wait is completed. Probably, I can remove the return in the case of error handling and have a common line to set the isRunning outside at line 59.

pkg/executor/kopia/kopiadelete.go
siva-portworx: Removed.

pkg/executor/kopia/kopiadelete.go
siva-portworx: Yes, correct. Removed the both the steps to check for the presence and creation.

pkg/drivers/kopiadelete/kopiadelete.go
```diff
@@ -0,0 +1,241 @@
...
+func (d Driver) DeleteJob(id string) error {
+	fn := "DeleteJob:"
+	namespace, name, err := utils.ParseJobID(id)
+	if err != nil {
+		logrus.Errorf("%s %v", fn, err)
+		return err
+	}
+
+	if err := coreops.Instance().DeleteSecret(name, namespace); err != nil && !apierrors.IsNotFound(err) {
+		errMsg := fmt.Sprintf("deletion of backup credential secret [%s] failed: %v", name, err)
```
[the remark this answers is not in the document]
siva-portworx: Fixed

pkg/drivers/kopiadelete/kopiadelete.go
```diff
@@ -0,0 +1,241 @@
...
+	if _, err = batch.Instance().CreateJob(job); err != nil && !apierrors.IsAlreadyExists(err) {
+		errMsg := fmt.Sprintf("creation of backup snapshot delete job [%s] failed: %v", jobName, err)
+		logrus.Errorf("%s %v", fn, errMsg)
+		return "", fmt.Errorf(errMsg)
+	}
+	logrus.Infof("%s created backup snapshot delete job [%s] successfully", fn, job.Name)
+	return utils.NamespacedName(job.Namespace, job.Name), nil
+}
+
+// DeleteJob stops backup snapshot delete job.
```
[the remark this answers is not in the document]
siva-portworx: Fixed

pkg/drivers/kopiadelete/kopiadelete.go
```diff
@@ -0,0 +1,241 @@
...
+	"github.com/sirupsen/logrus"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// SkipResourceAnnotation skipping kopia secret to be backed up
+	SkipResourceAnnotation = "stork.libopenstorage.org/skip-resource"
```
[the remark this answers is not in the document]
siva-portworx: Removed.

prashanthpx (approved):

PR #55, opened 2021-09-14
Changes to DataExport CR for the following

prashanthpx:
- Storing snapshotID from VolumeBackup CR into DatExport CR
- Creation of secret in source/destination ns for handling multi ns use case
  where stork backup/restore/BL CR's are present in kube-system
- Deletion of VolumeBackup CR after a successful backup

PB-1870

Manual Test
Updated DataExport CR

```
[root@prashanth-iron-seer-0 kdmp]# k describe dataexport kopia-aws-mysql-1 -nrst
Name:         kopia-aws-mysql-1
Namespace:    rst
Labels:       <none>
Annotations:  <none>
API Version:  kdmp.portworx.com/v1alpha1
Kind:         DataExport
Metadata:
  Creation Timestamp:  2021-09-14T15:05:31Z
  Finalizers:
    kdmp.portworx.com/finalizer-cleanup

Spec:
  Destination:
    API Version:  stork.libopenstorage.org/v1alpha1
    Kind:         BackupLocation
    Name:         kopia-bl
    Namespace:    kube-system
  Source:
    API Version:  v1
    Kind:         PersistentVolumeClaim
    Name:         mysql-data
    Namespace:    rst
  Type:           kopia
Status:
  Snapshot ID:  5cfcfd8453e76988c913a3daac3f8763
  Stage:        Final
  Status:       Successful
  Transfer ID:  rst/kopia-aws-mysql-1-mysql-data
Events:         <none>
```

### DIFF
--- a/pkg/drivers/kopiabackup/kopiabackup.go
+++ b/pkg/drivers/kopiabackup/kopiabackup.go
@@ -21,6 +21,7 @@ import (
 const (
 	// SkipResourceAnnotation skipping kopia secret to be backed up
 	SkipResourceAnnotation = "stork.libopenstorage.org/skip-resource"
+	backupJobPrefix        = "backup"
 )
 
 // Driver is a resticbackup implementation of the data export interface.
@@ -49,7 +50,7 @@ func (d Driver) StartJob(opts ...drivers.JobOption) (id string, err error) {
 	}
 	// DataExportName will be unique name when also generated from stork
 	// if there are multiple backups being triggered
-	jobName := toJobName(o.DataExportName, o.SourcePVCName)
+	jobName := toJobName(o.DataExportName)
 	logrus.Debugf("backup jobname: %s", jobName)
 	job, err := buildJob(jobName, o)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


pkg/drivers/kopiabackup/kopiabackup.go
```diff
@@ -49,7 +50,7 @@ func (d Driver) StartJob(opts ...drivers.JobOption) (id string, err error) {
 	}
 	// DataExportName will be unique name when also generated from stork
 	// if there are multiple backups being triggered
-	jobName := toJobName(o.DataExportName, o.SourcePVCName)
+	jobName := toJobName(o.DataExportName)
```
ram-infrac, line 53: can we also make volumebackup cr name consistent with dataexport cr

ram-infrac (approved):
lgtm

pkg/drivers/kopiabackup/kopiabackup.go
prashanthpx: The name is the same as DE CR but adding prefix string `backup` so that we know if VB CR is its created by offline method (kdmp separetely)

PR #59, opened 2021-09-22
PB-1802 Job framework

prabhu-px:
PB-1802 Job framework
Closes # PB-1802


pkg/drivers/utils/utils.go
```diff
@@ -73,6 +80,19 @@ func ToJobStatus(progress float64, errMsg string) *drivers.JobStatus {
...
+// GetConfigValue read configmap and return the value of the requested parameter
+func GetConfigValue(key string) string {
+	configMap, err := core.Instance().GetConfigMap(
+		kdmpConfig,
+		defaultPXNamespace,
+	)
+	if err != nil {
+		return ""
+	}
+	log.Errorf("Failed to read configmap.")
```
siva-portworx, line 92: Any reason why this error log is out the  if condition.

pkg/drivers/utils/utils.go
```diff
@@ -73,6 +80,19 @@ func ToJobStatus(progress float64, errMsg string) *drivers.JobStatus {
...
 }
 
+// GetConfigValue read configmap and return the value of the requested parameter
+func GetConfigValue(key string) string {
+	configMap, err := core.Instance().GetConfigMap(
+		kdmpConfig,
+		defaultPXNamespace,
+	)
+	if err != nil {
+		return ""
```
siva-portworx, line 90: Should we return some default value, if the entry for the given jobtype is not present in the configmap?
Empty string will be zero, that means the jobs will not started.

pkg/jobframework/jobframework.go
```diff
@@ -0,0 +1,49 @@
...
+	defaultNamespace = "kube-system"
+)
+
+// getJobsByType takes the jobType as a param and returns the list of jobs matching the label
+func getJobsByType(jobType string) *v1.JobList {
+	labelSelector := fmt.Sprintf("jobtype=%s", jobType)
+	options := metav1.ListOptions{
+		LabelSelector: labelSelector,
+	}
+	allJobs, _ := batch.Instance().ListAllJobs(defaultNamespace, options)
```
siva-portworx, line 22: Here we need to check jobs in all the namespace instead of kube-system alone

pkg/jobframework/jobframework.go
```diff
@@ -0,0 +1,49 @@
...
+		if job.Status.Active > 0 {
+			activeJobsCount++
+		}
+	}
+	return activeJobsCount
+}
+
+// jobLimitByType takes the job type and fetches the value from the config map
+func jobLimitByType(jobType string) int {
+	jobLimit, _ := strconv.Atoi(utils.GetConfigValue(jobType))
```
siva-portworx, line 39: It is better to handle error and in the case of error, it is good to assume some default value.

pkg/controllers/dataexport/reconcile.go
```diff
@@ -481,6 +482,9 @@ func (c *Controller) checkResticRestore(de *kdmpapi.DataExport) error {
 }
 
 func startTransferJob(drv drivers.Interface, srcPVCName string, dataExport *kdmpapi.DataExport) (string, error) {
+	if jobframework.JobCanRun(drv.Name()) == false {
```
siva-portworx, line 485: It it better to move this check below the drv == nil check.

pkg/controllers/dataexport/reconcile.go
```diff
@@ -485,6 +486,9 @@ func startTransferJob(drv drivers.Interface, srcPVCName string, dataExport *kdmp
 		return "", fmt.Errorf("data transfer driver is not set")
 	}
 
+	if jobframework.JobCanRun(drv.Name()) == false {
```
prashanthpx, line 489:
```suggestion
	if !jobframework.JobCanRun(drv.Name()) {
```

pkg/jobframework/jobframework.go
```diff
@@ -0,0 +1,65 @@
...
+				activeJobsCount++
+			}
+		}
+	}
+	return activeJobsCount
+}
+
+// jobLimitByType takes the job type and fetches the value from the config map
+func jobLimitByType(jobType string) int {
+	jobLimit, err := strconv.Atoi(utils.GetConfigValue(jobType))
```
prashanthpx, line 49: Any reason to pass empty string from `GetConfigValue()` and do Atoi? You could just check for empty string here and pass the default values or have default values passed from GetConfigValue() and consume here

prabhu-px: strconv.Atoi will fail in case of empty string as well as if by mistake config has invalid values . So i thought it can handle both at one condition

prashanthpx:
If the resource is not available when `startTransferJob()` fails then we need to set the status to `DataExportStatusInitial` or else currently it is set to `DataExportStatusFailed` for all failures and stork can mark this backup as failed by looking a this.

Also, you need to add labels to backup/restore/delete jobs.

Thinking through for snapshotDelete() from px-backup we are directly going to use job pkg and call `startJob()` in such case we can't rate limit delete jobs.  The initial thought of calling rate limit before `startTransferJob` might not hold good here.
We would need to think of a way to call this in `startJob()` itself and return the appropriate error msg

adityadani:
> If the resource is not available when `startTransferJob()` fails then we need to set the status to `DataExportStatusInitial` or else currently it is set to `DataExportStatusFailed` for all failures and stork can mark this backup as failed by looking a this.
> 
> Also, you need to add labels to backup/restore/delete jobs.
> 
> Thinking through for snapshotDelete() from px-backup we are directly going to use job pkg and call `startJob()` in such case we can't rate limit delete jobs. The initial thought of calling rate limit before `startTransferJob` might not hold good here.
> We would need to think of a way to call this in `startJob()` itself and return the appropriate error msg

I agree. This limit should be enforced at the job library instead of in the reconcilor.

siva-portworx:
This issue take care as part of https://github.com/portworx/kdmp/pull/79 PR. So closing it.

PR #60, opened 2021-09-23
PB-1882: Fixing issue with creating secret for multi ns

prashanthpx:
- Moved deleting VolumeBackup CR into cleanup() job
- Fixed issue in restore not updating VolumeBakcup CR for the failure scenario

PB-1882

### DIFF
--- a/pkg/drivers/kopiarestore/kopiarestore.go
+++ b/pkg/drivers/kopiarestore/kopiarestore.go
@@ -17,10 +17,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const (
-	restoreJobPrefix = "restore"
-)
-
 // Driver is a kopiarestore implementation of the data export interface.
 type Driver struct{}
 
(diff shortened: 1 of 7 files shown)


pkg/drivers/kopiarestore/kopiarestore.go
```diff
@@ -185,7 +188,7 @@ func jobFor(
 						{
 							Name:            "kopiaexecutor",
 							Image:           utils.KopiaExecutorImage(),
-							ImagePullPolicy: corev1.PullIfNotPresent,
+							ImagePullPolicy: corev1.PullAlways,
```
diptiranjanpx, line 191: Should we do PullAlways?

ram-infrac (approved):

pkg/drivers/kopiarestore/kopiarestore.go
prashanthpx: Till we do GA I feel this can be the option so that we can ask QA to pick up newbuilds (when we fix something) or else they have to delete images from all noes.  Before GA we can change this.

ram-infrac: we will be tagging  executor image for release, so it does not need to be change for GA as well. Let's keep it `PullAlways` to avoid going back & forth with QA for build fix

prashanthpx: But even if we have tagged, if this is PullAlways then for every backup triggered kopiaexecutor image would get pulled right which we want to avoid?

PR #70, opened 2021-10-01
PB-1921: Add support for passing custom certificates

prashanthpx:
Add supprot for passing custom certificates
- User creates secret out of public certificate, mounts to controller pod
- Provide certificate path in env SSL_CERT_DIR so job pods can pick and
  mount the certificate as part of jobs

PB-1921

Manual test
1. Deployed minio locally with self-signed certificates
2. Mounted the certificate as the secret to stork under kube-system ns
3. Add the following to the stork deployment spec
```
  - name: SSL_CERT_DIR
          value: "/etc/tls-secrets"

volumeMounts:
        - name: tls-secrets
          mountPath: "/etc/tls-secrets"
          readOnly: true

volumes:
      - name: tls-secrets
        secret:
          secretName: tls-s3-certs
```

4. Trigerred a generic backup/restore on proxy vol which was successful
5.

### DIFF
--- a/pkg/controllers/dataexport/reconcile.go
+++ b/pkg/controllers/dataexport/reconcile.go
@@ -3,7 +3,9 @@ package dataexport
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"reflect"
 	"strconv"
 
@@ -131,8 +133,22 @@ func (c *Controller) sync(ctx context.Context, in *kdmpapi.DataExport) (bool, er
 		if dataExport.Status.SnapshotPVCName != "" {
 			srcPVCName = dataExport.Status.SnapshotPVCName
 		}
-		// Create the credential secret
+
 		logrus.Debugf("drivername: %v", driverName)
+		// Creating a secret for ssl enabled objectstore with custom certificates
+		// User creates a secret in kube-system ns, mounts it to stork and recommendation is
+		// secret data is 'public.crt' always which contains the certificate content
+		// This secret path is provided as env section as follows
+		// env:
+		// - name: SSL_CERT_DIR
+		//   value: "/etc/tls-secrets" (can be any path)
+
+		// Check if the above env is present and read the certs file contents and
+		// secret for the job pod for kopia to access the same
+		err = createCertificateSecret(drivers.CertSecretName, dataExport.Spec.Source.Namespace)
+		if err != nil {
+			return false, err
+		}
 		if driverName == drivers.KopiaBackup {
 			// This will create a unique secret per PVC being backed up
 			// Create secret in source ns because in case of multi ns backup
@@ -148,6 +164,7 @@ func (c *Controller) sync(ctx context.Context, in *kdmpapi.DataExport) (bool, er
 				logrus.Errorf(msg)
 				return false, c.updateStatus(dataExport, kdmpapi.DataExportStatusFailed, msg)
 			}
+
 		}
 
 		if driverName == drivers.KopiaRestore {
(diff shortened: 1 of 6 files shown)


adityadani (approved):

pkg/controllers/dataexport/reconcile.go
```diff
@@ -232,7 +249,13 @@ func (c *Controller) sync(ctx context.Context, in *kdmpapi.DataExport) (bool, er
 		}
 		dataExport.Status.SnapshotID = volumeBackupCR.Status.SnapshotID
 		dataExport.Status.Size = volumeBackupCR.Status.TotalBytes
-
+		// Delete the tls certificate secret created
+		err = core.Instance().DeleteSecret(drivers.CertSecretName, dataExport.Spec.Source.Namespace)
+		if err != nil && errors.IsAlreadyExists(err) {
+			errMsg := fmt.Sprintf("failed to delete [%s:%s] secret", dataExport.Spec.Source.Namespace, drivers.CertSecretName)
```
siva-portworx, line 255: nit: It is better to add the err to be part of errMsg to know the actual error details.

pkg/drivers/kopiabackup/kopiabackup.go
```diff
@@ -229,7 +225,42 @@ func jobFor(
 				},
 			},
 		},
-	}, nil
+	}
+
+	if drivers.CertFilePath != "" {
```
siva-portworx, line 230: nit: Since update up to the job is done multiple, it is good to have it as utility function that takes job spec and update and return it.

pkg/controllers/dataexport/reconcile.go
```diff
@@ -123,8 +125,22 @@ func (c *Controller) sync(ctx context.Context, in *kdmpapi.DataExport) (bool, er
 		if dataExport.Status.SnapshotPVCName != "" {
 			srcPVCName = dataExport.Status.SnapshotPVCName
 		}
-		// Create the credential secret
+
 		logrus.Debugf("drivername: %v", driverName)
+		// Creating a secret for ssl enabled objectstore with custom certificates
```
siva-portworx, line 130: Query: How are we handing if more than one backuplocation is configured with the self-signed certificate?

pkg/drivers/kopiarestore/kopiarestore.go
```diff
@@ -230,7 +221,42 @@ func jobFor(
 				},
 			},
 		},
-	}, nil
+	}
+
```
siva-portworx, line 225: We might have to do this for other operation like delete and maintenance as well rt?

siva-portworx (approved):

pkg/controllers/dataexport/reconcile.go
```diff
@@ -232,7 +249,13 @@ func (c *Controller) sync(ctx context.Context, in *kdmpapi.DataExport) (bool, er
 		}
 		dataExport.Status.SnapshotID = volumeBackupCR.Status.SnapshotID
 		dataExport.Status.Size = volumeBackupCR.Status.TotalBytes
-
+		// Delete the tls certificate secret created
+		err = core.Instance().DeleteSecret(drivers.CertSecretName, dataExport.Spec.Source.Namespace)
+		if err != nil && errors.IsAlreadyExists(err) {
```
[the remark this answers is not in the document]
prashanthpx: Yeah my bad. Fixed it.

pkg/drivers/kopiabackup/kopiabackup.go
prashanthpx: Indeed i am planning to have single jobFor() and joForLive() clubbed into one.  During this cleanup, we can handle this also.

pkg/drivers/kopiarestore/kopiarestore.go
prashanthpx: Yeah, already have a ticket created for the same PB-1945

pkg/controllers/dataexport/reconcile.go
prashanthpx: Good point. As discussed, needs little brainstorming will create a ticket to track the to be addressed.

PR #96, opened 2021-11-01
pb-1988: Using the job secret and creating a temporary backuplocation for csi cleanup in KDMP controller

siva-portworx:
```
pb-1988: Using the job secret and creating a temporary backuplocation
    for csi cleanup in KDMP controller

        - Replaced client.Update to updateStatus which ha ve retry logic
          in kdmp controller
        - Added AlreadyExists and IsNotFound check while creating and
              deleting resources in the kdmp controller
        - Returning true from IsJobPending api, when the
          containerStatuses is empty.
```
Closes #
PB-1988

**Testing:**
1) Triggered a FA CSI backup with offload to S3 option.
2) Verified that all resources are deleted after backup.
3) Verified the latest CR copies are uploaded to the cloud as well.

### DIFF
--- a/pkg/controllers/dataexport/reconcile.go
+++ b/pkg/controllers/dataexport/reconcile.go
@@ -79,6 +79,9 @@ type updateDataExportDetail struct {
 	progressPercentage   int
 	snapshotPVCName      string
 	snapshotPVCNamespace string
+	snapshotNamespace    string
+	removeFinalizer      bool
+	volumeSnapshot       string
 }
 
 var volumeAPICallBackoff = wait.Backoff{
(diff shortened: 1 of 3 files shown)


pkg/controllers/dataexport/reconcile.go
```diff
@@ -611,14 +662,22 @@ func (c *Controller) stageSnapshotInProgress(ctx context.Context, dataExport *kd
 		return false, c.updateStatus(dataExport, data)
 	}
 
-	return true, c.client.Update(ctx, setStatus(dataExport, kdmpapi.DataExportStatusSuccessful, snapInfo.Reason))
+	data := updateDataExportDetail{
+		status: kdmpapi.DataExportStatusSuccessful,
+		reason: snapInfo.Reason,
```
diptiranjanpx, line 667: Do reason needs to be updated even if it is successful?

pkg/controllers/dataexport/reconcile.go
```diff
@@ -388,7 +414,7 @@ func (c *Controller) sync(ctx context.Context, in *kdmpapi.DataExport) (bool, er
 			return "", false, nil
 		}
 		if _, err := task.DoRetryWithTimeout(cleanupTask, defaultTimeout, progressCheckInterval); err != nil {
-			errMsg := fmt.Sprintf("max retries done, failed to delete [%s:%s] secret", dataExport.Spec.Source.Namespace, drivers.CertSecretName)
+			errMsg := fmt.Sprintf("max retries done, de: [%v/%v] cleanup failed with %v", dataExport.Namespace, dataExport.Name, err)
```
diptiranjanpx, line 417: Lets put "dataexport" instead of "de".

prashanthpx (approved):

pkg/controllers/dataexport/reconcile.go
siva-portworx: Fixed

pkg/controllers/dataexport/reconcile.go
siva-portworx: It was there previously, So added it. remove it now.

PR #97, opened 2021-11-02
[PB-1999]: For csi generic backup, kopia repo need to use the original source pvc instead of the staging pvc.

diptiranjanpx:
signed-off: Diptiranjan

Closes # PB-1999

**Test**:

1. Successful incremental generic backups for csi volumes with snapshotting.
2. Restoring backups.

### DIFF
--- a/pkg/controllers/dataexport/reconcile.go
+++ b/pkg/controllers/dataexport/reconcile.go
@@ -974,6 +974,7 @@ func startTransferJob(drv drivers.Interface, srcPVCName string, dataExport *kdmp
 	case drivers.KopiaBackup:
 		return drv.StartJob(
 			drivers.WithSourcePVC(srcPVCName),
+			drivers.WithRepoPVC(getRepoPVCName(dataExport, srcPVCName)),
 			drivers.WithNamespace(dataExport.Spec.Source.Namespace),
 			drivers.WithBackupLocationName(dataExport.Spec.Destination.Name),
 			drivers.WithBackupLocationNamespace(dataExport.Spec.Destination.Namespace),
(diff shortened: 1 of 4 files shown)


pkg/controllers/dataexport/reconcile.go
```diff
@@ -974,6 +974,7 @@ func startTransferJob(drv drivers.Interface, srcPVCName string, dataExport *kdmp
 	case drivers.KopiaBackup:
 		return drv.StartJob(
 			drivers.WithSourcePVC(srcPVCName),
```
prashanthpx, line 976: Do we still need this to be passed as it's not used anywhere?

pkg/drivers/options.go
```diff
@@ -180,6 +181,17 @@ func WithDestinationPVC(name string) JobOption {
 	}
 }
 
+// WithRepoPVC is job parameter.
+func WithRepoPVC(name string) JobOption {
+	return func(opts *JobOpts) error {
+		if strings.TrimSpace(name) == "" {
+			return fmt.Errorf("source pvc name should be set")
```
prashanthpx, line 188: Can we make this as `repo pvc name`?

pkg/controllers/dataexport/reconcile.go
```diff
@@ -1296,6 +1297,14 @@ func trimLabel(label string) string {
 	return label
 }
 
+func getRepoPVCName(de *kdmpapi.DataExport, pvcName string) string {
```
prashanthpx, line 1300: A thought as upfront we know this when calling `startTransferJob()`, can we `srcPVCName` itself with this logic being called before `startTransferJob()`?  That way changes could be minimal.

prashanthpx (approved):

pkg/controllers/dataexport/reconcile.go
diptiranjanpx: As discussed, the repo PVC name is just going to be used for the kopia repo in the cloud but the src PVC (in case of CSI volumes with snapshotting, it will be really the staging volume) will be used as it is.

PR #100, opened 2021-11-03
[STOR-551]: pvc creation logic is moved to kdmp reconciler in restore path.

diptiranjanpx:
signed-off-by: Diptiranjan

** Test**
Screenshot 2021-11-03 at 8 36 19 PM

### DIFF
--- a/pkg/controllers/dataexport/reconcile.go
+++ b/pkg/controllers/dataexport/reconcile.go
@@ -243,6 +243,31 @@ func (c *Controller) sync(ctx context.Context, in *kdmpapi.DataExport) (bool, er
 				}
 				return false, c.updateStatus(dataExport, data)
 			}
+
+			// Create the pvc from the spec provided in the dataexport CR
+			pvcSpec := dataExport.Status.RestorePVC
+			_, err = c.createPVC(dataExport)
+			if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controllers/dataexport/reconcile.go
```diff
@@ -994,6 +1019,18 @@ func (c *Controller) restoreSnapshot(ctx context.Context, snapshotDriver snapsho
 	return pvc, nil
 }
 
+func (c *Controller) createPVC(dataExport *kdmpapi.DataExport) (*corev1.PersistentVolumeClaim, error) {
+	pvc := dataExport.Status.RestorePVC
+	newPVC, err := core.Instance().CreatePersistentVolumeClaim(pvc)
+	if err != nil {
+		if k8sErrors.IsAlreadyExists(err) {
+			return pvc, nil
```
siva-portworx, line 1027: nit: It would be good, if you can return newPVC here

siva-portworx (approved):

pkg/controllers/dataexport/reconcile.go
diptiranjanpx: As discussed, followed the similar way of returning the input object in the case when it hits AlreadyExists error.

PR #109, opened 2021-11-10
Support for restore using local snapshots in generic backup

diptiranjanpx:
signed-off-by: Diptiranjan

If the local snapshot is available for a volume, try to do the restore from that.
if it is not available or restore fails, move to generic restore.

Closes # 
Screenshot 2021-11-10 at 6 13 56 PM
OC-513

### DIFF
--- a/pkg/controllers/dataexport/reconcile.go
+++ b/pkg/controllers/dataexport/reconcile.go
@@ -20,7 +20,6 @@ import (
 	"github.com/portworx/kdmp/pkg/drivers"
 	"github.com/portworx/kdmp/pkg/drivers/driversinstance"
 	"github.com/portworx/kdmp/pkg/drivers/utils"
-	"github.com/portworx/kdmp/pkg/snapshots"
 	kdmpopts "github.com/portworx/kdmp/pkg/util/ops"
 	"github.com/portworx/sched-ops/k8s/core"
 	"github.com/portworx/sched-ops/k8s/stork"
@@ -175,6 +174,10 @@ func (c *Controller) sync(ctx context.Context, in *kdmpapi.DataExport) (bool, er
 		return c.stageSnapshotRestore(ctx, dataExport)
 	case kdmpapi.DataExportStageSnapshotRestoreInProgress:
 		return c.stageSnapshotRestoreInProgress(ctx, dataExport)
+	case kdmpapi.DataExportStageLocalSnapshotRestore:
+		return c.stageLocalSnapshotRestore(ctx, dataExport)
+	case kdmpapi.DataExportStageLocalSnapshotRestoreInProgress:
+		return c.stageLocalSnapshotRestoreInProgress(ctx, dataExport)
 	case kdmpapi.DataExportStageTransferScheduled:
 		if dataExport.Status.Status == kdmpapi.DataExportStatusSuccessful {
 			// set to the next stage
@@ -474,7 +477,9 @@ func (c *Controller) stageInitial(ctx context.Context, dataExport *kdmpapi.DataE
 	if dataExport.Status.Status == kdmpapi.DataExportStatusSuccessful {
 		// set to the next stage
 		dataExport.Status.Stage = kdmpapi.DataExportStageTransferScheduled
-		if hasSnapshotStage(dataExport) {
+		if hasLocalRestoreStage(dataExport) {
+			dataExport.Status.Stage = kdmpapi.DataExportStageLocalSnapshotRestore
+		} else if hasSnapshotStage(dataExport) {
 			dataExport.Status.Stage = kdmpapi.DataExportStageSnapshotScheduled
 		}
 		data := updateDataExportDetail{
@@ -604,8 +609,7 @@ func (c *Controller) getSnapshotDriverName(dataExport *kdmpapi.DataExport) (stri
 	if err != nil && !k8sErrors.IsNotFound(err) {
 		return "", nil
 	}
-	// Default to external-storage snapshot class
-	return snapshots.ExternalStorage, nil
+	return "", fmt.Errorf("did not find any supported snapshot driver for snapshot storage class %s", dataExport.Spec.SnapshotStorageClass)
 }
 
 func (c *Controller) stageSnapshotInProgress(ctx context.Context, dataExport *kdmpapi.DataExport) (bool, error) {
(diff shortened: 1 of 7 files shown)


pkg/controllers/dataexport/reconcile.go
```diff
@@ -859,14 +862,205 @@ func getBackuplocationFromSecret(secretName, namespace string) (*storkapi.Backup
...
+		// set to the next stage
+		data := updateDataExportDetail{
+			stage:  kdmpapi.DataExportStageLocalSnapshotRestoreInProgress,
+			status: kdmpapi.DataExportStatusInitial,
+			reason: "",
+		}
+		return true, c.updateStatus(dataExport, data)
+	} else if dataExport.Status.Status == kdmpapi.DataExportStatusFailed {
+		err := c.cleanupLocalRestoredSnapshotResources(dataExport, false)
+		msg := fmt.Sprintf("failed in cleaning up resources restored for local snapshot restore for dataexport %s/%s: %v", dataExport.Namespace, dataExport.Name, err)
```
siva-portworx, line 876: msg initialization need to be inside the if err condition

pkg/controllers/dataexport/reconcile.go
```diff
@@ -859,14 +862,205 @@ func getBackuplocationFromSecret(secretName, namespace string) (*storkapi.Backup
...
+		if err != nil {
+			logrus.Debugf(msg)
+			data := updateDataExportDetail{
+				status: kdmpapi.DataExportStatusFailed,
+				reason: msg,
+			}
+			return false, c.updateStatus(dataExport, data)
+		}
+		data := updateDataExportDetail{
```
siva-portworx, line 885: It is good to have Informational message saying localsnap restore failed and switching to generic restore from cloud.

pkg/controllers/dataexport/reconcile.go
```diff
@@ -859,14 +862,205 @@ func getBackuplocationFromSecret(secretName, namespace string) (*storkapi.Backup
...
+		msg := fmt.Sprintf("Error while restoring from local snapshot with volumebackup %s in namespace %s : %v",
+			dataExport.Spec.Source.Name, dataExport.Spec.Source.Namespace, err)
+		data := updateDataExportDetail{
+			status: kdmpapi.DataExportStatusFailed,
+			reason: msg,
+		}
+		return false, c.updateStatus(dataExport, data)
+	}
+
+	if !status {
```
siva-portworx, line 950: Nit: the " if !status" can be combined with "if err != nil"

pkg/controllers/dataexport/reconcile.go
```diff
@@ -859,14 +862,205 @@ func getBackuplocationFromSecret(secretName, namespace string) (*storkapi.Backup
...
+		// set to the next stage
+		data := updateDataExportDetail{
+			stage:  kdmpapi.DataExportStageCleanup,
+			status: kdmpapi.DataExportStatusInitial,
+			reason: "",
+		}
+		return true, c.updateStatus(dataExport, data)
+	} else if dataExport.Status.Status == kdmpapi.DataExportStatusFailed {
+		err := c.cleanupLocalRestoredSnapshotResources(dataExport, false)
+		msg := fmt.Sprintf("failed in cleaning up resources restored for local snapshot restore for dataexport %s/%s: %v", dataExport.Namespace, dataExport.Name, err)
```
siva-portworx, line 981: msg initialization can be inside if err check

pkg/controllers/dataexport/reconcile.go
```diff
@@ -859,14 +862,205 @@ func getBackuplocationFromSecret(secretName, namespace string) (*storkapi.Backup
...
+	if err != nil {
+		msg := fmt.Sprintf("failed to get a snapshot restore status: %s", err)
+		data := updateDataExportDetail{
+			status: kdmpapi.DataExportStatusFailed,
+			reason: msg,
+		}
+		return false, c.updateStatus(dataExport, data)
+	}
+
+	if restoreInfo.Status == snapshotter.StatusFailed {
```
siva-portworx, line 1028: NIT: if err != nil and  if restoreInfo.Status == snapshotter.StatusFailed can be combined with || condition.

pkg/controllers/dataexport/reconcile.go
```diff
@@ -859,14 +862,205 @@ func getBackuplocationFromSecret(secretName, namespace string) (*storkapi.Backup
...
 func (c *Controller) cleanUp(driver drivers.Interface, de *kdmpapi.DataExport) error {
 	var bl *storkapi.BackupLocation
-
 	if driver == nil {
 		return fmt.Errorf("driver is nil")
 	}
-
-	if hasSnapshotStage(de) {
+	if hasLocalRestoreStage(de) {
+		err := c.cleanupLocalRestoredSnapshotResources(de, true)
```
prashanthpx, line 1058: When `cleanupLocalRestoredSnapshotResources()` is called in previous stages and as part of this PVC is deleted and stage moves to In-progress. Now as part of cleanup(), `cleanupLocalRestoredSnapshotResources` is called would that time deletion of PVC would fail as it's already deleted? Should this be idempotent?

vendor/github.com/libopenstorage/stork/pkg/snapshotter/snapshotter_csi.go
```diff
@@ -852,3 +877,112 @@ func (c *csiDriver) RetainLocalSnapshots(
...
+	var vsCRPath string
+
+	vsCRList, err := c.getCSISnapshotsCRList(backupLocation, pvcUID, objectPath)
+	if err != nil {
+		return snapshotInfo, fmt.Errorf("failed in getting list of older volumesnapshot CRs from objectstore : %v", err)
+	}
+	for _, volumeSnapshotCR := range vsCRList {
+		tempBackupUID, _ := getBackupInfoFromObjectKey(volumeSnapshotCR)
+		if tempBackupUID == backupUID {
+			logrus.Debugf("resources of local snapshot for backup %s are prsent in the objectpath %s", backupUID, objectPath)
```
prashanthpx, line 893: typo prsent

pkg/controllers/dataexport/reconcile.go
```diff
@@ -859,14 +862,205 @@ func getBackuplocationFromSecret(secretName, namespace string) (*storkapi.Backup
...
+			reason: "",
+		}
+		return true, c.updateStatus(dataExport, data)
+	} else if dataExport.Status.Status == kdmpapi.DataExportStatusFailed {
+		err := c.cleanupLocalRestoredSnapshotResources(dataExport, false)
+		msg := fmt.Sprintf("failed in cleaning up resources restored for local snapshot restore for dataexport %s/%s: %v", dataExport.Namespace, dataExport.Name, err)
+		if err != nil {
+			logrus.Debugf(msg)
+			data := updateDataExportDetail{
+				status: kdmpapi.DataExportStatusFailed,
```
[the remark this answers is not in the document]
siva-portworx: Proably, we can make the cleanupLocalRestoredSnapshotResources be called in the DoRetryWithTimeout and if it still fails, we can go ahead with cloud restore by moving to DataExportStageTransferScheduled stage.

pkg/controllers/dataexport/reconcile.go
```diff
@@ -474,7 +476,9 @@ func (c *Controller) stageInitial(ctx context.Context, dataExport *kdmpapi.DataE
 	if dataExport.Status.Status == kdmpapi.DataExportStatusSuccessful {
 		// set to the next stage
 		dataExport.Status.Stage = kdmpapi.DataExportStageTransferScheduled
-		if hasSnapshotStage(dataExport) {
+		if hasLocalRestoreStage(dataExport) {
+			dataExport.Status.Stage = kdmpapi.DataExportStageLocalSnapshotRestore
```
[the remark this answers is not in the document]
diptiranjanpx: when we are setting the flag for `hasLocalRestoreStage` to be true, that time "SnapshotStorageClass" is also set .
So i had added the first check to be `hasLocalRestoreStage` and in else if to check for `hasSnapshotStage `

vendor/github.com/libopenstorage/stork/pkg/snapshotter/snapshotter_csi.go
diptiranjanpx: Will take care of this in stork PR.

pkg/controllers/dataexport/reconcile.go
diptiranjanpx: Thats why the second flag which is true here is to skipPVC check. Because if the csi restore is successful as part of that cleanup we should not delete the PVC.

Coming to the case of idempotency wrt pvc, if ignorePVC is set to false and if pvc is already deleted, then that cleanup handles NotFound error.

pkg/controllers/dataexport/reconcile.go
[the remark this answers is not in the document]
diptiranjanpx: Addressed.

pkg/controllers/dataexport/reconcile.go
diptiranjanpx: Here the error messages are different. Hence keeping it same.

pkg/controllers/dataexport/reconcile.go
diptiranjanpx: done.

pkg/controllers/dataexport/reconcile.go
diptiranjanpx: done.

pkg/controllers/dataexport/reconcile.go
diptiranjanpx: done

pkg/controllers/dataexport/reconcile.go
diptiranjanpx: Here the error messages are different. Hence keeping it same.

prashanthpx (approved):
lgtm

PR #110, opened 2021-11-11
pb-2009: Add /tmp as default director for the kopia cache and log dir.

siva-portworx:
```
    pb-2009: Add /tmp as default director for the kopia cache and log dir.

        - Also added /tmp/kopiaconfig for default config that will be
          used. These changes are done as in some platform, we do not
          have permission to create dir or file under root dir.
        - In the case of retry returning actual error instead of
          task.DoRetryWithTimeout function error.
```
Closes # pb-2009

Tested it on the OCS setup.

### DIFF
--- a/pkg/kopia/command.go
+++ b/pkg/kopia/command.go
@@ -8,7 +8,10 @@ import (
 )
 
 const (
-	baseCmd = "kopia"
+	baseCmd    = "kopia"
+	logDir     = "/tmp"
+	cacheDir   = "/tmp"
+	configFile = "/tmp/kopiconfig"
 )
 
 // Command defines the essential fields required to
@@ -88,6 +91,12 @@ func (c *Command) CreateCmd() *exec.Cmd {
 			c.Password,
 			"--prefix",
 			c.RepositoryName,
+			"--log-dir",
+			logDir,
+			"--cache-directory",
+			cacheDir,
+			"--config-file",
+			configFile,
 		}
 	case "s3":
 		argsSlice = []string{
@@ -100,6 +109,12 @@ func (c *Command) CreateCmd() *exec.Cmd {
 			c.Password,
 			"--prefix",
 			c.RepositoryName,
+			"--cache-directory",
+			cacheDir,
+			"--log-dir",
+			logDir,
+			"--config-file",
+			configFile,
 		}
 
 		if c.DisableSsl {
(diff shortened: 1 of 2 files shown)


pkg/kopia/command.go
```diff
@@ -88,6 +91,12 @@ func (c *Command) CreateCmd() *exec.Cmd {
 			c.Password,
 			"--prefix",
 			c.RepositoryName,
+			"--log-dir",
+			"logDir",
```
diptiranjanpx, line 95: these are variables , right? not fixed strings.

siva-portworx: Oops correct. I missed it. Thanks.

diptiranjanpx (approved):

pkg/kopia/command.go
siva-portworx: fixed

pkg/kopia/command.go
```diff
@@ -88,6 +91,12 @@ func (c *Command) CreateCmd() *exec.Cmd {
 			c.Password,
 			"--prefix",
 			c.RepositoryName,
+			"--log-dir",
+			logDir,
+			"--cache-directory",
```
prashanthpx, line 96: As this is common, could have been added here argsSlice = append(argsSlice, c.Flags...).

pkg/kopia/command.go
```diff
@@ -140,6 +161,10 @@ func (c *Command) BackupCmd() *exec.Cmd {
 	argsSlice := []string{
 		"snapshot",
 		c.Name, // create command
+		"--log-dir",
+		logDir,
```
prashanthpx, line 165: It's not needed for backupcmd and as we are already passing it when the repository is created.
Same for restore command.
In general, we need to add for the following only
1. Repo create
2. Repo connect
Every other command uses repo create so it should be taken care of.

siva-portworx: As discussed, while testing the log-dir permission issue seen even when the repo connect went through the log-dir option. So added it for all the command. Only the cache-dir option is valid for the repo create and repo connect.

pkg/kopia/command.go
siva-portworx: Followed like other options (config and password). Will take up the cleanup in next PR.

PR #111, opened 2021-11-11
Handle default snapshot classes

adityadani:
Treat a snapshot class name provided as "default" or "Default" specifically for CSI snapshot provider.
- If such a snapshot class is detected then assume that it is the CSI snapshotter provider.

Closes #

### DIFF
--- a/pkg/controllers/dataexport/reconcile.go
+++ b/pkg/controllers/dataexport/reconcile.go
@@ -587,6 +587,10 @@ func (c *Controller) getSnapshotDriverName(dataExport *kdmpapi.DataExport) (stri
 	if len(dataExport.Spec.SnapshotStorageClass) == 0 {
 		return "", fmt.Errorf("snapshot storage class not provided")
 	}
+	if dataExport.Spec.SnapshotStorageClass == "default" ||
+		dataExport.Spec.SnapshotStorageClass == "Default" {
+		return csiProvider, nil
+	}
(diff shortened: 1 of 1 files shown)


prashanthpx (approved):

pkg/controllers/dataexport/reconcile.go
```diff
@@ -587,6 +587,10 @@ func (c *Controller) getSnapshotDriverName(dataExport *kdmpapi.DataExport) (stri
 	if len(dataExport.Spec.SnapshotStorageClass) == 0 {
 		return "", fmt.Errorf("snapshot storage class not provided")
 	}
+	if dataExport.Spec.SnapshotStorageClass == "default" ||
+		dataExport.Spec.SnapshotStorageClass == "Default" {
+		return csiProvider, nil
+	}
```
siva-portworx, line 593: Query: If we do not set the volumesnapshot class, will the K8s will automatically pick the one that was set as default?

siva-portworx (approved):

pkg/controllers/dataexport/reconcile.go
adityadani: Yes k8s will select the default one. And if the admin has not set any snapshot class as the default one, then the snapshot operation will fail. Now this is exactly what we want. The user requested to use the "default" snapshot class, and there is none set so we should fail the snapshot operation and thus the backup operation.

PR #112, opened 2021-11-15
PB-2023: Add skipResourceAnnotation for intermediate PVC created

prashanthpx:
- As part of CSI + KDMP backup, intermediate PVC is created, corresponding
  PV for the same needs to be skipped being backed up

PB-2023

Manual test
1. Trigerred an all ns backup having
2. Checked the PV and PVC count are matching
Attached screenshot for reference
Screenshot 2021-11-16 at 12 29 18 AM (2)


pkg/controllers/dataexport/reconcile.go
```diff
@@ -812,6 +812,23 @@ func (c *Controller) stageSnapshotRestoreInProgress(ctx context.Context, dataExp
 		status: kdmpapi.DataExportStatusSuccessful,
 		reason: restoreInfo.Reason,
 	}
+	// Add annotation for corresponding PV to be skipped backing
+	tempName := toSnapshotPVCName(srcPvc.Name, string(dataExport.UID))
+	interPvc, err := core.Instance().GetPersistentVolumeClaim(tempName, srcPvc.Namespace)
+	if err != nil {
```
diptiranjanpx, line 818: is the idea is to ignore the error?

diptiranjanpx (approved):

PR #119, opened 2021-11-18
PB-2042: Fixed backup failure retry due when fatalerror count is encountered

prashanthpx:
```
  PB-2042: Fixed backup failure retry due when fatalerror count is encountered
    - As part of a reconciler, before the error is reported in stage In-progress, the reconciler will check
      if the job has exhausted pod retires amount to backof limit.
    - If the pod has restarted this many times, the job will be marked as failed and moved to
      cleanup stage.
```

PB-2042

Manual test
1. Simulated error by pointing a wrong backuploation
2. Trigerred a backup
3.  Checked the job pod was restarted till backoff limit is reached and no resources were deleted
4.  After exhausting backoff limit, all resources were deleted and backup marked as failure.
5

### DIFF
--- a/pkg/executor/kopia/kopiabackup.go
+++ b/pkg/executor/kopia/kopiabackup.go
@@ -530,24 +530,51 @@ func isRepositoryExists(repository *executor.Repository) (bool, error) {
 		logrus.Errorf("%v", err)
 		return false, err
 	}
-	bucket, err := objectstore.GetBucket(bl)
-	if err != nil {
-		logrus.Errorf("err: %v", err)
-		return false, err
+	exists := false
+	t := func() (interface{}, bool, error) {
+		bucket, err := objectstore.GetBucket(bl)
+		if err != nil {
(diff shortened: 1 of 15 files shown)


pkg/executor/kopia/kopiabackup.go
```diff
@@ -530,24 +530,51 @@ func isRepositoryExists(repository *executor.Repository) (bool, error) {
...
+		}
+
+		return "", false, nil
 	}
-	if exists {
-		logrus.Infof("%s exists", kopiaRepositoryFile)
-	} else {
-		logrus.Infof("%s doesn't exists", kopiaRepositoryFile)
+
+	if _, err := task.DoRetryWithTimeout(t, executor.DefaultTimeout, progressCheckInterval); err != nil {
```
siva-portworx, line 570: Can get the actual error from the retry steps and print. error of DoRetryWithTimeout is not useful

siva-portworx (approved):

PR #125, opened 2021-11-23
PB-2076: Additional check if node exists in the JobStatus call.

diptiranjanpx:
signed-off-by: Diptiranjan

If the job is running and the node on which the job pod is scheduled gets shut down,
the pod keeps on getting spawned and the corresponding operation (like backup) will be in that stage forever.

With this PR, there is one more additional check in JobStatus has been introduced to check if the node exists or not.

Closes # Pb-2076

Test:
1. The already running job moved to the Failed state (backup - pb-21-system-test-57-ela-2f78453-interval-2021-11-22-131324)
2. Took a new successful backup (backup - pb-21-system-test-57-ela-2f78453-interval-2021-11-23-073026)
3. Again while backup was going on, poweroff one more node (backup - pb-21-system-test-57-ela-2f78453-interval-2021-11-23-081506 and pb-21-system-test-57-ela-2f78453-interval-2021-11-23-081506)

Screenshot 2021-11-23 at 2 05 21 PM

### DIFF
--- a/pkg/drivers/utils/utils.go
+++ b/pkg/drivers/utils/utils.go
@@ -113,6 +113,31 @@ func IsJobPending(j *batchv1.Job) bool {
 	return false
 }
 
+// JobNodeExists checks if the node associated with the job spec exists in the cluster
+func JobNodeExists(j *batchv1.Job) error {
+	// Get the node name from job spec.
+	nodeName := j.Spec.Template.Spec.NodeName
+	if len(nodeName) > 0 {
+		nodes, err := core.Instance().GetNodes()
+		if err != nil {
+			logrus.Errorf("getting all nodes failed with error: %v", err)
+			// Not returning the error, so that the job does not get marked as failed in case of flaky errors
+			return nil
+		}
+		found := false
+		for _, node := range nodes.Items {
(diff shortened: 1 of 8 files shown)


prashanthpx (approved):

pkg/drivers/utils/utils.go
```diff
@@ -113,6 +113,29 @@ func IsJobPending(j *batchv1.Job) bool {
...
 }
 
+// JobNodeExists checks if the node associated with the job spec exists in the cluster
+func JobNodeExists(j *batchv1.Job) error {
+	// Get the node name from job spec.
+	nodeName := j.Spec.Template.Spec.NodeName
+	if len(nodeName) > 0 {
+		nodes, err := core.Instance().GetNodes()
+		if err != nil {
+			return err
```
siva-portworx, line 123: Instead of returning error, log the error message and return nil. Let us not fail, if we can not get the node list.

siva-portworx (approved):

pkg/drivers/utils/utils.go
diptiranjanpx: done

PR #127, opened 2021-11-24
PB-2075: Making steps as 14 so that timeout happens in 20-30 mins range.

diptiranjanpx:
signed-off-by: Diptiranjan

Reducing the timeout for volume bound check 

Closes # PB-2075


prashanthpx (approved):

siva-portworx (approved):
LGTM

PR #129, opened 2021-12-08
[PB-2084]: Reading dataexport CR with schedops api instead of contoller client call.

diptiranjanpx:
signed-off-by: Diptiranjan

Even after updating dataexport CR, for some of the fields we used to get stale entries.
Hence the change to move to sched ops api.

Closes # PB-2084

Test : Updated in PB-2084


siva-portworx (approved):
LGTM

PR #130, opened 2021-12-10
pb-2113: Fixed the invalid access to the storage class variable if it is not present in the PV spec.

siva-portworx:
pb-2113: Fixed the invalid access to the storage class variable if it is not present in the PV spec.

Closes # PR-2113

Tested on the QA setup. Now it is not crashing.

### DIFF
--- a/pkg/controllers/dataexport/reconcile.go
+++ b/pkg/controllers/dataexport/reconcile.go
@@ -1518,16 +1518,22 @@ func checkPVCIgnoringJobMounts(in kdmpapi.DataExportObjectReference, expectedMou
 		if checkErr != nil {
 			return "", true, checkErr
 		}
-		var sc *storagev1.StorageClass
 		storageClassName := k8shelper.GetPersistentVolumeClaimClass(pvc)
 		if storageClassName != "" {
-			sc, checkErr = storage.Instance().GetStorageClass(storageClassName)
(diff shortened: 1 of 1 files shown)


diptiranjanpx (approved):

pkg/controllers/dataexport/reconcile.go
```diff
@@ -1526,8 +1526,15 @@ func checkPVCIgnoringJobMounts(in kdmpapi.DataExportObjectReference, expectedMou
 				return "", true, checkErr
 			}
 			logrus.Debugf("checkPVCIgnoringJobMounts: pvc name %v - storage class VolumeBindingMode %v", pvc.Name, *sc.VolumeBindingMode)
-		}
-		if *sc.VolumeBindingMode != storagev1.VolumeBindingWaitForFirstConsumer {
+			if *sc.VolumeBindingMode != storagev1.VolumeBindingWaitForFirstConsumer {
```
diptiranjanpx, line 1529: can remove the declaring of that "sc" variable before getting the storageclass.

prashanthpx (approved):

pkg/controllers/dataexport/reconcile.go
siva-portworx: Fixed.

PR #131, opened 2021-12-11
pb-2113: cleanup in error handling in waitForPVCBound and checkPVCIgnoringJobMounts functions.

siva-portworx:
```
pb-2113: cleanup in error handling in waitForPVCBound and checkPVCIgnoringJobMounts functions.
```
Closes #pb-2113

**Testing:**
**With out Fix:**
Screenshot 2021-12-11 at 10 32 06 PM

**With Fix:**

Screenshot 2021-12-11 at 10 32 47 PM

### DIFF
--- a/pkg/controllers/dataexport/reconcile.go
+++ b/pkg/controllers/dataexport/reconcile.go
@@ -1485,6 +1485,7 @@ func waitForPVCBound(in kdmpapi.DataExportObjectReference, checkMounts bool) (*c
 	// wait for pvc to get bound
 	var pvc *corev1.PersistentVolumeClaim
 	var err error
+	var errMsg string
 	wErr := wait.ExponentialBackoff(volumeAPICallBackoff, func() (bool, error) {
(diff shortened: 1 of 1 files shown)


pkg/controllers/dataexport/reconcile.go
```diff
@@ -1502,7 +1503,7 @@ func waitForPVCBound(in kdmpapi.DataExportObjectReference, checkMounts bool) (*c
 
 	if wErr != nil {
 		logrus.Errorf("%v", wErr)
-		return nil, wErr
+		return nil, err
```
diptiranjanpx, line 1506: can we return the error message also in wErr , such that from the final message it will be "timeout happened with the %v error), err"

prashanthpx (approved):
lgtm

pkg/controllers/dataexport/reconcile.go
siva-portworx: Fixed.

PR #135, opened 2021-12-16
pb-2131: Fixed couple of error handling issues.

siva-portworx:
```
pb-2131: Fixed couple of error handling issues.

        - if snapshotDriveName returned by getSnapshotDriverName is empty, ignore it in cleanupLocalRestoredSnapshotResources
          function and move on to kdmp restore stage.
        - In other instance, if snapshotDriveName returned by
          getSnapshotDriverName is empty, consider it as error.
        - The final error cleanupErr return as part of
          cleanupLocalRestoredSnapshotResources function is overwritten
          and getting nil error, in the case of error. Fixed it.
```
Closes # pb-2131

Testing:
Tested restore on the setup that did not has the volumesnapshot class. It defaulted to S3 object restore.
Note:
1) Currently the reason is getting updated to restore of volume successful. The reason "switching to restore from objectstore bucket as restoring from local snapshot did not happen" is getting updated after success. Will discuss on the option.

### DIFF
--- a/pkg/controllers/dataexport/reconcile.go
+++ b/pkg/controllers/dataexport/reconcile.go
@@ -467,7 +467,7 @@ func (c *Controller) sync(ctx context.Context, in *kdmpapi.DataExport) (bool, er
 		cleanupTask := func() (interface{}, bool, error) {
 			cleanupErr := c.cleanUp(driver, dataExport)
 			if cleanupErr != nil {
-				errMsg := fmt.Sprintf("failed to remove resources: %s", err)
+				errMsg := fmt.Sprintf("failed to remove resources: %s", cleanupErr)
 				logrus.Errorf("%v", errMsg)
 				return "", true, fmt.Errorf("%v", errMsg)
 			}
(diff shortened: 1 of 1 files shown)


diptiranjanpx (approved):

pkg/controllers/dataexport/reconcile.go
```diff
@@ -627,7 +627,8 @@ func (c *Controller) getSnapshotDriverName(dataExport *kdmpapi.DataExport) (stri
 	if err != nil && !k8sErrors.IsNotFound(err) {
 		return "", nil
 	}
-	return "", fmt.Errorf("did not find any supported snapshot driver for snapshot storage class %s", dataExport.Spec.SnapshotStorageClass)
+	logrus.Warnf("did not find any supported snapshot driver for snapshot storage class %s", dataExport.Spec.SnapshotStorageClass)
```
prashanthpx, line 630: The old logic of returning errors was fine right? Also in line#628 we need to return the error. With both these changes, there would be no need to check for an empty snapshot name in the caller.
Also in the caller now when snapshot name is empty (in some cases), this error msg `fmt.Errorf("failed to get snapshot driver name: %v", err)` will have an empty error and might be confusing to the user.

prashanthpx (approved):

prashanthpx (approved):
lgtm

pkg/controllers/dataexport/reconcile.go
siva-portworx: Fixed

PR #138, opened 2022-01-08
pb-2156: Failing the backup, if the local snapshot backup or restore fails

siva-portworx:
pb-2156: Failing the backup, if the local snapshot backup or restore fails.

Closes # pb-2156

Testing:
Tested on OCP cluster by commenting out the default to KDMP in the case of the OCP provisioner.
After timeout, backup will be marked with the failure instead of retry and hanging.

Screenshot 2022-01-08 at 4 36 26 PM

### DIFF
--- a/pkg/controllers/dataexport/reconcile.go
+++ b/pkg/controllers/dataexport/reconcile.go
@@ -569,6 +569,25 @@ func (c *Controller) stageSnapshotScheduled(ctx context.Context, dataExport *kdm
 	if err != nil {
 		return false, fmt.Errorf("failed to get snapshot driver for %v: %v", snapshotDriverName, err)
 	}
+	// This will create a unique secret per PVC being backed up
+	// Create secret in source ns because in case of multi ns backup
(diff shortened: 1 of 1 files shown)


prashanthpx (approved):

pkg/controllers/dataexport/reconcile.go
```diff
@@ -648,6 +667,15 @@ func (c *Controller) stageSnapshotInProgress(ctx context.Context, dataExport *kd
 		}
 		return true, c.updateStatus(dataExport, data)
 	}
+	if dataExport.Status.Status == kdmpapi.DataExportStatusFailed {
+		// set to the next stage
+		data := updateDataExportDetail{
+			stage:  kdmpapi.DataExportStageCleanup,
+			status: dataExport.Status.Status,
+			reason: "",
```
diptiranjanpx, line 675: lets keep the reason in both the failed cases.

siva-portworx: Fixed.

PR #140, opened 2022-01-14
[PB-2148]: support for cleaning up of binding jobs used for restoring volumes with binding mode as waitforfirstconsumers.

diptiranjanpx:
signed-off-by: Diptiranjan
cleaning up of dataexport crs should not depend on volumebackup secret.
storage class annotation needs to be kept in staging pvc if source pvc already has it.

cleaning up of binding jobs used for making pvc bound.
clean up of dataexport cr needs to use the precreated secret for backuplocation info
if source pvc has storage class annotation set, use that for staging pvc creation.

Closes # PB-2148

**Test**
Covered following scenarios

Generic Backup and restore of csi volumes with snapshotting with volumebinding mode as WaitForFirstConsumer.
Generic Backup and restore of csi volumes with snapshotting with volumebinding mode as immediate.
Native backup and restore of gke volumes with snapshotting with volumebinding mode as WaitForFirstConsumer.

### DIFF
--- a/pkg/controllers/dataexport/reconcile.go
+++ b/pkg/controllers/dataexport/reconcile.go
@@ -21,6 +21,7 @@ import (
 	"github.com/portworx/kdmp/pkg/drivers/driversinstance"
 	"github.com/portworx/kdmp/pkg/drivers/utils"
 	kdmpopts "github.com/portworx/kdmp/pkg/util/ops"
+	"github.com/portworx/sched-ops/k8s/batch"
 	"github.com/portworx/sched-ops/k8s/core"
 	"github.com/portworx/sched-ops/k8s/storage"
 	"github.com/portworx/sched-ops/k8s/stork"
@@ -51,6 +52,7 @@ const (
 	dataExportUIDAnnotation  = "portworx.io/dataexport-uid"
 	dataExportNameAnnotation = "portworx.io/dataexport-name"
 	skipResourceAnnotation   = "stork.libopenstorage.org/skip-resource"
+	baseSCAnnotation         = "volume.beta.kubernetes.io/storage-class"
 
 	pxbackupAnnotationPrefix        = "portworx.io/"
 	kdmpAnnotationPrefix            = "kdmp.portworx.com/"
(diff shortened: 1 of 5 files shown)


pkg/controllers/dataexport/reconcile.go
```diff
@@ -1165,6 +1189,9 @@ func (c *Controller) cleanUp(driver drivers.Interface, de *kdmpapi.DataExport) e
 			}
 		}
 		if de.Status.SnapshotPVCName != "" && de.Status.SnapshotPVCNamespace != "" {
+			if err := cleanupJobBoundResources(de.Status.SnapshotPVCName, de.Status.SnapshotPVCNamespace); err != nil {
```
prashanthpx, line 1192: Was thinking if we also want to move this cleanup to stork cleanup of kdmp resources as stork is the one which creates the job.  
Add comments to know where the job resource is created for this cleanup.

prashanthpx (approved):

pkg/controllers/dataexport/reconcile.go
diptiranjanpx: the cleanup which has been added in stork_csi.go is really to handle the success path where the volume becomes bound . these are the cleanups which are required if the pvc does not get bound.

pkg/controllers/dataexport/reconcile.go
```diff
@@ -64,7 +66,10 @@ const (
 
 	// pvcNameLenLimit is the max length of PVC name that DataExport related CRs
 	// will incorporate in their names
-	pvcNameLenLimit       = 247
+	pvcNameLenLimit = 247
+	// pvcNameLenLimitForJob is the max length of PVC name that the bound job
+	// will incorporate in their names
+	pvcNameLenLimitForJob = 237
```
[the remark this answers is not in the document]
diptiranjanpx: The job name can be max 253 chars.
And the name is something like fmt.Sprintf("%s-%s-%s", "bound", truncatedPVCName, uidToken[0])

so need to truncate the pvc name if required to the pvcNameLenLimitForJob

PR #142, opened 2022-03-06
pb-2196: Added check such that add hostaccess and privileged access only for live backup.

siva-portworx:
```
pb-2196: Added check such that add hostaccess and privileged access only for live backup.

        - Also creating the job related resources in the kube-system for the
          live backup
                - service account
                - tls certificate secret
                - credential secret
                - volumebackup
                - job
                - job pods
        - Fixed the issue of TLS certificate secret getting created in the
          incorrect namespace for restore operation.
        - Also referring to dataexport.Namespace as a common namespace both the
          restore and backup secret creation.
```
Closes # pb-2196

Testing:
1) Deployed a mysql with one PVC with proxy volume.
2) Made replica to zero to test the non live backup.
3) Made replica to zero to test live backup
4) Made sure that all the job resources created in the kube-system for the live backup is getting deleted.
5) Also checked the roles content between the live and non-live backup. Attached the screen shot for reference.
Screenshot 2022-03-06 at 8 40 10 AM
Screenshot 2022-03-06 at 9 50 53 AM

### DIFF
--- a/pkg/controllers/dataexport/reconcile.go
+++ b/pkg/controllers/dataexport/reconcile.go
@@ -212,46 +212,10 @@ func (c *Controller) sync(ctx context.Context, in *kdmpapi.DataExport) (bool, er
 		}
 
 		logrus.Debugf("drivername: %v", driverName)
-		// Creating a secret for ssl enabled objectstore with custom certificates
-		// User creates a secret in kube-system ns, mounts it to stork and recommendation is
-		// secret data is 'public.crt' always which contains the certificate content
-		// This secret path is provided as env section as follows
-		// env:
-		// - name: SSL_CERT_DIR
-		//   value: "/etc/tls-secrets" (can be any path)
-
-		// Check if the above env is present and read the certs file contents and
-		// secret for the job pod for kopia to access the same
-		err = createCertificateSecret(drivers.CertSecretName, dataExport.Spec.Source.Namespace, dataExport.Labels)
-		if err != nil {
-			return false, err
-		}
-		if driverName == drivers.KopiaBackup {
-			// This will create a unique secret per PVC being backed up
-			// Create secret in source ns because in case of multi ns backup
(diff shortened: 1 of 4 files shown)


pkg/controllers/dataexport/reconcile.go
```diff
@@ -500,6 +450,78 @@ func (c *Controller) sync(ctx context.Context, in *kdmpapi.DataExport) (bool, er
...
+) (updateDataExportDetail, error) {
+	namespace := dataExport.Namespace
+	// Below are default values for kopiaRestore. There values will be updated with different, if the driveName is kopiaBackup.
+	var blName, blNamespace string
+	if driverName == drivers.KopiaRestore {
+		blName = vb.Spec.BackupLocation.Name
+		blNamespace = vb.Spec.BackupLocation.Namespace
+	}
+	if driverName == drivers.KopiaBackup {
+		pods, err := core.Instance().GetPodsUsingPVC(srcPVCName, dataExport.Spec.Source.Namespace)
```
prashanthpx, line 467: As I know there would be only one kopia pod always per PVC (we don't allow to take simultaneous backup on the same PVC in parallel), under what conditions would these lengths be > 1? 

Wouldn't this also get other pods (like app) ones that are using the PVC?

siva-portworx: This is to check whether the PVC that we are planning to take snapshot is already used by some application. In that case, we will be taking live backup path. For this case, the Job spec will be created in the kube-system namespace. So all the dependant resources like secret, TLS secret, service account and Volumebackup will be created in the kube-system.

siva-portworx: One issue is there is that TLS cert secret is always created with hardcoded name, it will conflict. Similar to other resource, I will use dataexport.Name for this as well.

prashanthpx (approved):

PR #155, opened 2022-04-15
pb-2293: add logic to extract registry that contains extra directories in it to use for kopia executor image

siva-portworx:
```
pb-2293: add logic to extract registry that contains extra directories in it to use for kopia executor image
```
Closes # pb-2293

United tested with different combination of image format.


px-kesavan:
@siva-portworx Please change the title to pb-2292

px-kesavan (approved):

PR #159, opened 2022-04-23
pb-2330: Made fix to support image name to have custom repo as well.

siva-portworx:
```
 pb-2330: Made fix to support image name to have custom repo as well.
```
Closes # pb-2330

```
Now we will all the below image format:
 <custom-registry-name>/<extra-dir-name>/<custom-repo-name>/image:tag
 <custom-registry-name>/<custom-repo-name>/<extra-dir-name>/image:tag
 <repo-name>/image:tag -- default registry of docker.io will be taken.
 ```

### DIFF
--- a/pkg/drivers/utils/utils.go
+++ b/pkg/drivers/utils/utils.go
@@ -235,10 +235,9 @@ func GetImageRegistryFromDeployment(name, namespace string) (string, string, err
 	}
 	imageFields := strings.Split(deploy.Spec.Template.Spec.Containers[0].Image, "/")
 	// Here the assumption is that the image format will be <registry-name>/<extra-dir-name>/<repo-name>/image:tag
-	// or <repo-name>/image:tag. If repo name contains any path (<registry-name>/<repo-name>/<extra-dir-name>/image:tag), below logic will not work.
+	// or <repo-name>/image:tag or <registry-name>/<repo-name>/<extra-dir-name>/image:tag).
 	// Customer might have extra dirs before the repo-name as mentioned above
-	// here minus 2 is for image name and repo name exclusion
-	registryFields := imageFields[0 : len(imageFields)-2]
+	registryFields := imageFields[0 : len(imageFields)-1]
 	registry := strings.Join(registryFields, "/")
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by prashanthpx — APPROVED** [review]

File: pkg/drivers/utils/utils.go [hunk 1]
```diff
@@ -235,10 +235,9 @@ func GetImageRegistryFromDeployment(name, namespace string) (string, string, err
 	}
 	imageFields := strings.Split(deploy.Spec.Template.Spec.Containers[0].Image, "/")
 	// Here the assumption is that the image format will be <registry-name>/<extra-dir-name>/<repo-name>/image:tag
-	// or <repo-name>/image:tag. If repo name contains any path (<registry-name>/<repo-name>/<extra-dir-name>/image:tag), below logic will not work.
+	// or <repo-name>/image:tag or <registry-name>/<repo-name>/<extra-dir-name>/image:tag).
 	// Customer might have extra dirs before the repo-name as mentioned above
-	// here minus 2 is for image name and repo name exclusion
-	registryFields := imageFields[0 : len(imageFields)-2]
+	registryFields := imageFields[0 : len(imageFields)-1]
```
adityadani, line 240: When no registry is provided and the image name is just `openstorage/stork:2.10`. In that case we would end up with a kopia executor image as : `openstorage/portworx/kopiaexecutor:<version>`. Isn't that right?

siva-portworx: Sorry I did not noticed the review comment and merged it. 

I will modify the KopiaExecutorImage to just kopiaexecutor alone. (https://github.com/portworx/kdmp/blob/2eb295a0aa48507133ebc143239185031d0f12d1/pkg/drivers/drivers.go#L22)

But one caveat is that they need to upload the kopiaexecutor to different repos, if the px-backup and stork repo names are different. If the they have same custom repo name for all the images, then it is fine. 

I think it is ok to ask them to have kopiaexecutor image to present in both the px-backup and stork repo separetly, if they have different repos for them.
Is that fine? @adityadani

PR #163, opened 2022-05-18
pb-2328: Support both v1 & v1beta1 version of the volumeSnapshot

lalat-das:
- Added support for VolumeSnapshot V1 Apis which is GA since kubernetes
  1.20 version.
- The v1beta1 is still kept for supporting older K8s versions

**What this PR does / why we need it**: The volumesnapshot API v1beta1 support is removed from k8s version 1.20, hence we need to add v1 APIs along with v1beta1. 

Closes # pb-2328

### DIFF
--- a/pkg/controllers/dataexport/reconcile.go
+++ b/pkg/controllers/dataexport/reconcile.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	kSnapshotv1 "github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1"
 	kSnapshotv1beta1 "github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1beta1"
 	kSnapshotClient "github.com/kubernetes-csi/external-snapshotter/client/v4/clientset/versioned"
 	storkapi "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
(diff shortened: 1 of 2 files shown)


pkg/controllers/dataexport/reconcile.go
```diff
@@ -769,18 +771,38 @@ func (c *Controller) stageSnapshotInProgress(ctx context.Context, dataExport *kd
...
+		err = snapshotDriver.UploadSnapshotObjects(bl, snapInfoList, getCSICRUploadDirectory(pvcUID), getVSFileName(backupUID, timestampEpoch))
+		if err != nil {
+			msg := fmt.Sprintf("uploading snapshot objects for pvc %s/%s failed with error : %v", vs.Namespace, vs.Name, err)
+			logrus.Errorf(msg)
+			data := updateDataExportDetail{
+				status: kdmpapi.DataExportStatusFailed,
+				reason: msg,
+			}
+			return false, c.updateStatus(dataExport, data)
 		}
```
px-kesavan, line 805: This block of code is redundant, as it is also used in v1SnapshotRequired condition. Can we have a specific error message for v1 and v1beta ?. Else we can bring the error check out of if..else

px-kesavan: I overlooked on the variables that is being printed. if those are interfaces it makes sense to be in their corresponding conditional blocks.

px-kesavan (approved):
LGTM

pkg/controllers/dataexport/reconcile.go
lalat-das: Done the changes..

PR #177, opened 2022-09-07
pb-3091: Enable kopia repo maintenance for NFS BL

lalat-das:
- Added the NFS volume mount specific entries in JOB spec
- identify the kopia repo on the NFS mount path for running the kopia maintenance command.

**What this PR does / why we need it**: Enables the kopia maintenance job to run on NFS based backuplocation

Closes # pb-3091

Still Unit testing the changes

### DIFF
--- a/pkg/executor/kopia/maintenance.go
+++ b/pkg/executor/kopia/maintenance.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"time"
 
+	storkapi "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
 	"github.com/libopenstorage/stork/pkg/objectstore"
 	kdmp_api "github.com/portworx/kdmp/pkg/apis/kdmp/v1alpha1"
 	"github.com/portworx/kdmp/pkg/executor"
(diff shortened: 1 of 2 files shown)


pkg/executor/kopia/maintenance.go
```diff
@@ -213,6 +238,19 @@ func runMaintenance(maintenanceType string) error {
...
 
+func returnDirList(parentDir string) ([]string, error) {
+	var files []string
+	fileInfo, err := ioutil.ReadDir(parentDir)
+	if err != nil {
+		return files, err
+	}
+
+	for _, file := range fileInfo {
+		files = append(files, file.Name())
```
px-kesavan, line 249:
```suggestion
		if file.IsDir() {
			files = append(files, file.Name())
		}
```
Return only the directories and not the files inside it

pkg/executor/kopia/maintenance.go
```diff
@@ -213,6 +238,19 @@ func runMaintenance(maintenanceType string) error {
...
+	}
+	return files, nil
```
px-kesavan, line 251: nit : Can we have different naming, as function says `returnDirList`, but we are returning list named `files`

pkg/executor/kopia/maintenance.go
lalat-das: @kesavan the ReadDir API will not return file entries rather it will return directory entries all the time under the directory that is passed to the call. hence this check wont be necessary . Moreover we only create these directories under generic backup hence there won't be a situation to get a file entry here.

pkg/executor/kopia/maintenance.go
lalat-das: In unix/Linux  everything is a file.. :-)

lalat-das: This function actually returns only directory entries..

PR #196, opened 2022-10-18
Pb 3146 : kopia maintenance job should bail out properly for resource only backup scenario

lalat-das:
**What this PR does / why we need it**: This make the maintenance JOB to complete with an appropriate warning message when user takes a resource only backup. 

Closes # pb-3146

Before fix:
Pasted Graphic
Pasted Graphic 1

After fix:
image


lalat-das:
Wrongly raised against master..

PR #197, opened 2022-10-18
pb-3146: For resource only backup maintenance job should pass

lalat-das:
When there is a resource only backup the kopia maintenace job goes to crashloopbackoff state and keeps restarting. This is avoided by adding appropriate error check.

**What this PR does / why we need it**: For resource only backup the maintenance job crashloopbackoff contineously due to unavailability of generic-backup dir. Hence added appropriate error check to bail out gracefully.

Closes # pb-3146

Before fix :
Pasted Graphic

Pasted Graphic 1
After fix:
image

### DIFF
--- a/pkg/executor/kopia/maintenance.go
+++ b/pkg/executor/kopia/maintenance.go
@@ -136,7 +136,11 @@ func runMaintenance(maintenanceType string) error {
 		repoBaseDir := repo.Path + genericBackupDir + "/"
 		listOfSubDirs, err := returnDirList(repoBaseDir)
 		if err != nil {
-			logrus.Errorf("Failed to list sub directories in dir %v", repoBaseDir)
+			if os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


pkg/executor/kopia/maintenance.go
```diff
@@ -136,7 +136,11 @@ func runMaintenance(maintenanceType string) error {
 		repoBaseDir := repo.Path + genericBackupDir + "/"
 		listOfSubDirs, err := returnDirList(repoBaseDir)
 		if err != nil {
-			logrus.Errorf("Failed to list sub directories in dir %v", repoBaseDir)
+			if os.IsNotExist(err) {
+				logrus.Warnf("There are no %v directory exist, verify the backup unless it is a resource only backup", repoBaseDir)
```
diptiranjanpx, line 140: Nit: "No directory %v exists, verify if it is a resource only backup", repoBaseDir

PR #201, opened 2022-10-27
pb-3126: Added support for nfs volume restore

prashanthpx:
```
- NFS executor changes for vol restore
- ResourceExport and ResourceBackup CR to carry vols to be restored information
- NFS executor creates PVC to be restored as a separate job
```

PB-3126


pkg/executor/nfs/nfsrestorevolcreate.go
```diff
@@ -0,0 +1,511 @@
...
+	// Check if all the PVC's are in bounded state
+	rb, err := kdmpschedops.Instance().GetResourceBackup(rbCrName, rbCrNamespace)
+	if err != nil {
+		errMsg := fmt.Sprintf("error reading ResourceBackup CR[%v/%v]: %v", rbCrNamespace, rbCrName, err)
+		return fmt.Errorf(errMsg)
+	}
+	// Updating vol info
+
+	rb.VolumesInfo = resourceBackupVolInfos
+	rb.ExistingVolumesInfo = existingResourceBackupVolInfos
```
siva-portworx, line 249: volumeInfo update is done in-memory copy and it is not push to the CR. I could see a call to update ResourceBackup upfront itself at line 234.

siva-portworx (approved):

PR #202, opened 2022-10-30
pb-3159 : Added changes for sub-path & mount option for volume backup

lalat-das:
**What this PR does / why we need it**: This will add support for mountoption passing while taking backup on a NFS Share based backuplocation 

Closes # pb-3159

Yet to add changes to all other places where job is created and some optimisation in PV pvc cleanup path.

### DIFF
--- a/pkg/controllers/dataexport/reconcile.go
+++ b/pkg/controllers/dataexport/reconcile.go
@@ -37,7 +37,6 @@ import (
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/client-go/rest"
 	k8shelper "k8s.io/component-helpers/storage/volume"
@@ -74,9 +73,7 @@ const (
 	// pvcNameLenLimitForJob is the max length of PVC name that the bound job
 	// will incorporate in their names
 	pvcNameLenLimitForJob = 48
-	volumeinitialDelay    = 2 * time.Second
-	volumeFactor          = 1.5
-	volumeSteps           = 15
+
 	defaultTimeout        = 1 * time.Minute
 	progressCheckInterval = 5 * time.Second
 	compressionKey        = "KDMP_COMPRESSION"
@@ -96,12 +93,6 @@ type updateDataExportDetail struct {
 	snapshotNamespace    string
 	removeFinalizer      bool
 	volumeSnapshot       string
-}
-
-var volumeAPICallBackoff = wait.Backoff{
-	Duration: volumeinitialDelay,
-	Factor:   volumeFactor,
-	Steps:    volumeSteps,
 }
 
 func (c *Controller) sync(ctx context.Context, in *kdmpapi.DataExport) (bool, error) {
@@ -304,6 +295,7 @@ func (c *Controller) sync(ctx context.Context, in *kdmpapi.DataExport) (bool, er
 			utils.KdmpConfigmapNamespace,
 			backupLocation.Location.NfsConfig.ServerAddr,
 			backupLocation.Location.Path,
+			backupLocation.Location.NfsConfig.MountOption,
 		)
 		if err != nil && err != utils.ErrJobAlreadyRunning && err != utils.ErrOutOfJobResources {
 			msg := fmt.Sprintf("failed to start a data transfer job, dataexport [%v]: %v", dataExport.Name, err)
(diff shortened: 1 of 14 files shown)


lalat-das:
All test passing, I have added code to delete sa and secret in resource export reconciler as part of this PR. We can merge this PR and vendor to stork and px-backup

pkg/executor/common.go
```diff
@@ -290,7 +292,7 @@ func ParseCloudCred() (*Repository, error) {
 	}
 	if storkapi.BackupLocationType(blType) == storkapi.BackupLocationNFS {
 		// For NFS this path need to be absolute path not just a bucket name anymore.
-		repository.Path = drivers.NfsMount
+		repository.Path = drivers.NfsMount + repository.NfsConfig.SubPath
```
siva-portworx: Do we expect subPath to start with "/"? Is it not better that we add it, if it is missing?

pkg/drivers/utils/utils.go
```diff
@@ -641,3 +651,135 @@ func GetImageSecretName(name string) string {
...
+			Resources: corev1.ResourceRequirements{
+				Requests: corev1.ResourceList{
+					corev1.ResourceName(corev1.ResourceStorage): resource.MustParse("3Mi"),
+				},
+			},
+			VolumeName: pvName,
+		},
+	}
+
+	// Let's wait for PV creation and timeout after certain timeouts.
```
siva-portworx, line 722: Suggestion: It is better to wait for PV in the createNfsPV itself.. instead of doing it in PVC creation.

pkg/controllers/dataexport/reconcile.go
```diff
@@ -1332,6 +1324,33 @@ func (c *Controller) cleanUp(driver drivers.Interface, de *kdmpapi.DataExport) e
 		if err != nil && !k8sErrors.IsNotFound(err) {
 			return fmt.Errorf("delete %s job: %s", de.Status.TransferID, err)
 		}
+		//TODO : Need better way to find BL type from de CR
+		// For now deleting unconditionally for all BL type.
+
+		// secret, err := core.Instance().GetSecret(de.Name, de.Spec.Source.Namespace)
```
siva-portworx, line 1330: Can we remove the comment code and have one TODO comment alone.
I feel, we can go ahead with the approach of delete them unconditionally and if it is NotFound, we can ignore the error and continue.

pkg/controllers/resourceexport/reconcile.go
```diff
@@ -230,6 +231,22 @@ func (c *Controller) cleanupResources(resourceExport *kdmpapi.ResourceExport) er
...
 	}
+	pvcName := "pvc-" + rbName
+	if err := core.Instance().DeletePersistentVolumeClaim(pvcName, rbNamespace); err != nil && !k8sErrors.IsNotFound(err) {
+		return fmt.Errorf("delete %s/%s pvc: %s", rbNamespace, pvcName, err)
+	}
+	pvName := "pv-" + rbName
+	if err := core.Instance().DeletePersistentVolume(pvName); err != nil && !k8sErrors.IsNotFound(err) {
+		return fmt.Errorf("delete %s pv: %s", pvName, err)
+	}
+	if err := core.Instance().DeleteServiceAccount(rbName, rbNamespace); err != nil && !k8sErrors.IsNotFound(err) {
```
siva-portworx, line 242: Should we also take care of role/rolebinding cleanup?

pkg/drivers/kopiadelete/kopiadelete.go
```diff
@@ -84,6 +84,23 @@ func (d Driver) StartJob(opts ...drivers.JobOption) (id string, err error) {
 		logrus.Errorf("%s %v", fn, errMsg)
 		return "", fmt.Errorf(errMsg)
 	}
+
+	// Create PV & PVC only in case of NFS.
+	if o.NfsServer != "" {
```
siva-portworx, line 89: Suggestion: Can move these step to a util function as it repeated in all the jobs

pkg/drivers/utils/utils.go
```diff
@@ -641,3 +651,135 @@ func GetImageSecretName(name string) string {
...
+	pv := &corev1.PersistentVolume{
+		TypeMeta: metav1.TypeMeta{Kind: "PersistentVolume"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: pvName,
+			Annotations: map[string]string{
+				SkipResourceAnnotation: "true",
+			},
+		},
+		Spec: corev1.PersistentVolumeSpec{
+			//VolumeMode: v1.PersistentVolumeMode(),
```
siva-portworx, line 672: Remove commented code

siva-portworx (approved):

pkg/executor/common.go
lalat-das: NfsMount variable is already added with a "/". Moreover I have also added a "/" at the end

pkg/drivers/utils/utils.go
lalat-das: Done. Created a new function called WaitForPVavaialble() and called from PV creation path.

pkg/controllers/dataexport/reconcile.go
lalat-das: Sure Will do

pkg/drivers/kopiadelete/kopiadelete.go
lalat-das: Done

pkg/controllers/resourceexport/reconcile.go
lalat-das: yeah we should. Taken care of the same ..

pkg/drivers/utils/utils.go
lalat-das: done

pkg/drivers/utils/utils.go
```diff
@@ -644,3 +656,169 @@ func GetImageSecretName(name string) string {
...
+		return err
+	}
+
+	return nil
+}
+
+// CreateNFSPvPvcForJob - this function creates PV and PVC for NFS job.
+func CreateNFSPvPvcForJob(jobName string, namespace string, o drivers.JobOpts) error {
+	// Create PV & PVC only in case of NFS.
+	if o.NfsServer != "" {
```
px-kesavan, line 750: Why we require this additional check ?. The function name looks like its only for nfs

lalat-das: good catch that came because of a copy paste work while fixing review comments.. Will chnage and push again

PR #221, opened 2022-11-12
pb-3276: Updating the unstructed PVC object to point to correct namespace based on the ns mapping.

siva-portworx:
```
pb-3276: Updating the unstructed PVC object to point to correct namespace based on the ns mapping.
```
Closes # pb-3276

Verified following two cases:
Case1:
- Took single namespace backup and deleted the source SC
- Triggered restore to different namespace and different SC.

Case2:
- Took multi namespace backup of same app ( to have same PVC name)
- Deleted the source SC.
- Maunally created the destination namespaces ( as fix for creation of multi namepspace is pending)
- Trigged restore to different namespace and different SC.

### DIFF
--- a/pkg/executor/nfs/nfsrestorevolcreate.go
+++ b/pkg/executor/nfs/nfsrestorevolcreate.go
@@ -301,7 +301,7 @@ func restoreVolResourcesAndApply(
 			}
 			// TODO: In future we can optimize this path where we can directly get runtime.Unstructured from the preRestoreObjects
 			// for the respective pvc
-			o, err := fetchObjectFromPVC(preRestoreObjects, pvc)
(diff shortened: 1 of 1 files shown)


pkg/executor/nfs/nfsrestorevolcreate.go
```diff
@@ -363,7 +364,19 @@ func fetchObjectFromPVC(
 			if err != nil {
 				return nil, err
 			}
-			if tempPVC.Name == pvc.Name {
+			restoreNamespace, ok := namespaceMapping[tempPVC.Namespace]
+			if !ok {
+				return nil, fmt.Errorf("sivakumar restore namespace mapping not found: %s namespaceMapping %v ", restoreNamespace, namespaceMapping)
```
prashanthpx, line 369: Pls remove `sivakumar`

PR #223, opened 2022-11-13
pb-3278 Added support for creating namespare if not present during vol restore

prashanthpx:
`Added support for creating namespace if not present during vol restore`

pb-3278

ResourceExport CR created in kube-system ns
ResourceBackup CR created in kube-system ns
PVC creation happens in individual ns

Backup screen shot of two NS 
Screenshot 2022-11-13 at 10 59 06 PM (2)

Restore of the above (to two ns)
Screenshot 2022-11-13 at 10 59 06 PM (2)

Restoring to one namesapce, backup taken for two ns
Screenshot 2022-11-13 at 10 31 58 PM (2)

Restoring with retain option for existing ns
Screenshot 2022-11-14 at 6 53 02 AM (2)

### DIFF
--- a/pkg/executor/common.go
+++ b/pkg/executor/common.go
@@ -2,6 +2,7 @@ package executor
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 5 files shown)


pkg/executor/common.go
```diff
@@ -584,3 +602,158 @@ func getShortUID(uid string) string {
...
+		logrus.Errorf("%s: error parsing cloud cred: %v", funct, err)
+		return err
+	}
+	bkpDir := filepath.Join(repo.Path, backup.Status.BackupPath)
+	restoreLocation, err := storkops.Instance().GetBackupLocation(backup.Spec.BackupLocation, backupLocationNamespace)
+	if err != nil {
+		return err
+	}
+
+	nsData, err := DownloadObject(bkpDir, NamespacesFile, restoreLocation.Location.EncryptionKey)
```
siva-portworx, line 642: We need to use EncryptionV2Key instead of EncryptionKey

siva-portworx (approved):

PR #224, opened 2022-11-14
pb-3216: Taking QPS and burst value from the kdmp-config for nfs executors

siva-portworx:
```
 pb-3216: Taking QPS and burst value from the kdmp-config for nfs executors
```
Closes # pb-3216

### DIFF
--- a/pkg/drivers/utils/common.go
+++ b/pkg/drivers/utils/common.go
@@ -39,6 +39,14 @@ const (
 	KdmpConfigmapNamespace = "kube-system"
 	// DefaultCompresion default compression type
 	DefaultCompresion = "s2-parallel-8"
+	// DefaultQPS - default qps value for k8s apis
+	DefaultQPS = 100
+	// DefaultBurst - default burst value for k8s apis
+	DefaultBurst = 100
+	// QPSKey - configmap QPS key name
+	QPSKey = "K8S_QPS"
+	// BurstKey - configmap burst key name
+	BurstKey = "K8S_BURST"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


pkg/drivers/utils/common.go
```diff
@@ -39,6 +39,14 @@ const (
 	KdmpConfigmapNamespace = "kube-system"
 	// DefaultCompresion default compression type
 	DefaultCompresion = "s2-parallel-8"
+	// DefaultQPS - default qps value for k8s apis
+	DefaultQPS = 100
+	// DefaultBurst - default burst value for k8s apis
+	DefaultBurst = 100
+	// QPSKey - configmap QPS key name
+	QPSKey = "KDMP_QPS"
```
prashanthpx, line 47: How about making this as K8S_QPS? Same for below

pkg/executor/nfs/nfsbkpresources.go
```diff
@@ -457,10 +458,28 @@ func min(x, y int) int {
 }
 
 func initResourceCollector() resourcecollector.ResourceCollector {
+	QPS := kdmputils.DefaultQPS
+	Burst := kdmputils.DefaultBurst
+	kdmpData, err := core.Instance().GetConfigMap(kdmputils.KdmpConfigmapName, kdmputils.KdmpConfigmapNamespace)
+	if err != nil {
+		logrus.Warnf("failed reading config map %v: %v", kdmputils.KdmpConfigmapName, err)
+		logrus.Warnf("default to %s compression", kdmputils.DefaultCompresion)
```
prashanthpx, line 466: What is this compression referring to ?

prashanthpx (approved):

pkg/drivers/utils/common.go
siva-portworx: Done

pkg/executor/nfs/nfsbkpresources.go
siva-portworx: Cut and paste error. Fixed it.

PR #231, opened 2022-11-20
pb-3317: Avoiding the pod count for the pvc usage for live backup calculation, if it is used by kdmp job pods.

siva-portworx:
```
pb-3317: Avoiding the pod count for the pvc usage for live backup
    calculation, if it is used by kdmp job pods.
```
Closes # pb-3317

Tested with four parallel backup o
Screenshot 2022-11-20 at 1 11 38 PM
n same namespace.

### DIFF
--- a/pkg/controllers/dataexport/reconcile.go
+++ b/pkg/controllers/dataexport/reconcile.go
@@ -540,7 +540,15 @@ func (c *Controller) createJobCredCertSecrets(
 			}
 			return data, err
 		}
-		if len(pods) > 0 {
+		// filter out the pods that are create by us
+		count := len(pods)
+		for _, pod := range pods {
+			labels := pod.ObjectMeta.Labels
+			if _, ok := labels[drivers.DriverNameLabel]; ok {
+				count--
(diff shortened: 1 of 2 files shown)


pkg/controllers/dataexport/reconcile.go
```diff
@@ -540,7 +540,15 @@ func (c *Controller) createJobCredCertSecrets(
 			}
 			return data, err
 		}
-		if len(pods) > 0 {
+		// filter out the pods that are create by us
+		count := len(pods)
+		for _, pod := range pods {
+			labels := pod.ObjectMeta.Labels
+			if _, ok := labels[drivers.DriverNameLabel]; ok {
+				count = count - 1
```
diptiranjanpx, line 548: nit: count--

diptiranjanpx (approved):

pkg/controllers/dataexport/reconcile.go
siva-portworx: Fixed

PR #248, opened 2023-02-16
PB-3453: Uploading csi snapshot info for generic backups.

diptiranjanpx:
Signed-Off-By: Diptiranjan

Uploading csi snapshot info for volumes which have taken csi snapshot path for generic backup.

Closes # PB-3453

### DIFF
--- a/pkg/executor/nfs/nfsbkpresources.go
+++ b/pkg/executor/nfs/nfsbkpresources.go
@@ -48,6 +48,7 @@ const (
 	resourcesFile             = "resources.json"
 	storageClassFile          = "storageclass.json"
 	backupResourcesBatchCount = 15
+	volumeSnapShotCRDirectory = "csi-generic"
 )
 
 func newUploadBkpResourceCommand() *cobra.Command {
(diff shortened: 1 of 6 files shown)


pkg/executor/nfs/nfsbkpresources.go
```diff
@@ -162,6 +163,20 @@ func uploadBkpResource(
 		return fmt.Errorf(errMsg)
 	}
 
+	csiGenericBackupDirectory := filepath.Join(repo.Path, volumeSnapShotCRDirectory)
```
siva-portworx, line 166: Should add a check such that both uploadCSISnapshots and uploadCSISnapshotInfoForPVCs will be called always.
I mean uploadCSISnapshots should called only for native CSI and uploadCSISnapshotInfoForPVCs for kdmp case.

siva-portworx: ok ignore. We are checking for driverName inside the function.

siva-portworx (approved):

PR #249, opened 2023-03-04
pb-3637: Added S3 region option to the kopia command

siva-portworx:
```
pb-3637: Added S3 region option to the kopia command
```
Closes #pb-3637

Tested with the Oracle S3 objectstore with different region value.

### DIFF
--- a/pkg/executor/kopia/kopiabackup.go
+++ b/pkg/executor/kopia/kopiabackup.go
@@ -215,6 +215,7 @@ func runKopiaCreateRepo(repository *executor.Repository) error {
 			repository.Name,
 			repository.Password,
 			kopiaProviderType[repository.Type],
+			repository.S3Config.Region,
 			repository.S3Config.DisableSSL,
 		)
 	default:
@@ -223,6 +224,7 @@ func runKopiaCreateRepo(repository *executor.Repository) error {
 			repository.Name,
 			repository.Password,
 			kopiaProviderType[repository.Type],
+			"",
 			false,
 		)
 	}
@@ -350,6 +352,7 @@ func runKopiaRepositoryConnect(repository *executor.Repository) error {
 			repository.Name,
 			repository.Password,
 			kopiaProviderType[repository.Type],
+			repository.S3Config.Region,
 			repository.S3Config.DisableSSL,
 		)
 	default:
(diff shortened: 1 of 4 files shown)


pkg/executor/kopia/kopiabackup.go
```diff
@@ -223,6 +224,7 @@ func runKopiaCreateRepo(repository *executor.Repository) error {
 			repository.Name,
 			repository.Password,
 			kopiaProviderType[repository.Type],
+			"",
```
prashanthpx, line 227: You might want to not hav this for NFS region option at all. Can we check what is the imapcy on NFS vols?

siva-portworx: It is not an issue as we are not passing the --region option for the nfs ( non S3 bl target).
I check the kopia command between S3 and nfs. The output is as follow:
```
S3 oracle:
time="2023-03-04T19:54:55Z" level=info msg="sivakumar --- Connect Command /usr/bin/kopia repository connect s3 --bucket px-backup --password k0p1@#GB --prefix generic-backup/mysql-pso-mysql-data/ --cache-directory /tmp --log-dir /tmp --config-file /tmp/kopiaconfig --region eu-frankfurt-1 --endpoint axnfm4jb3i73.compat.objectstorage.eu-frankfurt-1.oraclecloud.com:443 --access-key 74208679a898479fa09474f874889ebc7f538ed9 --secret-access-key 5uBFf7tdR6TFKJBc3e9fqGHwtOqK0XCubcc4n2fjcXw="

NFS:
time="2023-03-04T20:04:59Z" level=info msg="sivakumar --- Create Command /usr/bin/kopia repository create filesystem --path /tmp/nfs-target//generic-backup/mysql-pso-mysql-data/ --password k0p1@#GB --cache-directory /tmp --log-dir /tmp --config-file /tmp/kopiaconfig"
time="2023-03-04T20:05:04Z" level=info msg="sivakumar --- Connect Command /usr/bin/kopia repository connect filesystem --path /tmp/nfs-target//generic-backup/mysql-pso-mysql-data/ --password k0p1@#GB --cache-directory /tmp --log-dir /tmp --config-file /tmp/kopiaconfig"
```
Screenshot 2023-03-05 at 1 45 44 AM

diptiranjanpx (approved):

pkg/executor/kopia/kopiabackup.go
prashanthpx: Yup i see it. Ideally i feel we should not even fil in command struct with region if not needed. The way we are taking case is stuct has all parameters and we choose in command section what needs to be added.

prashanthpx (approved):

PR #250, opened 2023-03-07
pb-3633: Added v1/v1beta version check in calling volumesnapshot class get in getSnapshotDriverName()

siva-portworx:
```
pb-3633: Added v1/v1beta version check in calling volumesnapshot class get in getSnapshotDriverName()
```
Closes # pb-3633

Tested on the QA setup, Now we are not seeing the deprecated message.


lalat-das (approved):
lgtm

PR #265, opened 2023-05-08
Pb 3842 large resource support in nfs

lalat-das:
**What this PR does / why we need it**: This adds support for LargeResource support for NFS backup location

Closes # pb-3842

Unit test : 
image

show status of large resource restore 
image

largeResource falg enabled in case of restore 
image

### DIFF
--- a/pkg/apis/kdmp/v1alpha1/resourcebackup.go
+++ b/pkg/apis/kdmp/v1alpha1/resourcebackup.go
@@ -35,17 +35,17 @@ const (
 )
 
 // ResourceApplyPhase defines a status of resource restoration phase.
-type ResourceApplyPhase string
+type ResourceApplyStatus string
 
 const (
 	// ResourceApplyPhasePreparing signifies we are altering the resource
 	// as per target cluster before applying
-	ResourceApplyPhasePreparing ResourceApplyPhase = "Preparing"
+	ResourceApplyPhasePreparing ResourceApplyStatus = "Preparing"
 	// ResourceApplyPhaseDeleting when Resources are deleted before applying this
 	// applicable only when replace is opted during restore process.
-	ResourceApplyPhaseDeleting ResourceApplyPhase = "Deleting"
+	ResourceApplyPhaseDeleting ResourceApplyStatus = "Deleting"
 	// ResourceApplyPhaseApplying when Resource is being applied as part of restore process
-	ResourceApplyPhaseApplying ResourceApplyPhase = "Applying"
+	ResourceApplyPhaseApplying ResourceApplyStatus = "Applying"
 )
 
 // ResourceBackupProgressStatus overall resource backup/restore progress
@@ -58,12 +58,12 @@ type ResourceBackupProgressStatus struct {
 	Reason string `json:"reason,omitempty"`
 	// Resources status of each resource being restore
 	Resources []*ResourceRestoreResourceInfo `json:"resources"`
-	// Total Resource Count is the total number of resources being restored
+	// TotalResourceCount is the total number of resources being restored
 	TotalResourceCount int64
-	// Restored Resource count is the count of already applied resource count
+	// RestoredResourceCount is the count of already applied resource count
 	RestoredResourceCount int64
-	// Resource apply phase signifies the current stage of resource restore
-	ResourceApplyStage ResourceApplyPhase
+	// ResourceApplyStage signifies the current stage of resource restore
+	ResourceApplyStage ResourceApplyStatus
 	// LargeResourceEnabled signifies if it largeResource enabled or not
 	LargeResourceEnabled bool
 }
--- a/pkg/apis/kdmp/v1alpha1/resourcebackup.go
+++ b/pkg/apis/kdmp/v1alpha1/resourcebackup.go
@@ -34,7 +34,7 @@ const (
 	ResourceBackupStatusPartialSuccess ResourceBackupStatus = "PartialSuccess"
 )
 
-// ResourceBackupStatus defines a status of ResourceBackup.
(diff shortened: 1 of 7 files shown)


pkg/executor/nfs/nfsrestoreresources.go
```diff
@@ -548,6 +564,12 @@ func applyResources(
 			RancherProjectMappings: rancherProjectMapping,
 		}
 	}
+
+	rb.Status.ResourceApplyStage = kdmpapi.ResourceApplyPhasePreparing
+	rb, err = executor.UpdateStatusInResourceBackup(rb.Status, rb.Name, rb.Namespace)
+	if err != nil {
+		logrus.Errorf("failed to update resorucebackup[%v/%v] status after hitting error in create namespace : %v", rbCrNamespace, rbCrName, err)
```
prashanthpx, line 571: Shouldn't we return from err on error or else we would not be seeing status in RB CR but silently in the background resoruces are applied. My thinking is, would be difficult for debugging also.

pkg/executor/nfs/nfsrestoreresources.go
```diff
@@ -614,33 +662,67 @@ func applyResources(
...
+				"Resource restored successfully", tempResourceList,
+			); err != nil {
 				return err
 			}
 		}
 	}
+	// append the temp slice to the final restore resouce list,
+	rb.Status.Resources = append(rb.Status.Resources, tempResourceList...)
+	rb.Status.RestoredResourceCount = int64(len(rb.Status.Resources))
 	// rb.Status.Reason = utils.ResourceUploadSuccessMsg
```
prashanthpx, line 696: nit: Remove commented code

pkg/executor/nfs/nfsrestoreresources.go
```diff
@@ -614,33 +662,67 @@ func applyResources(
...
+		if err != nil {
+			log.ApplicationRestoreLog(restore).Errorf("failed to read config map %v's key %v, setting default value of 1MB", k8sutils.StorkControllerConfigMapName,
+				k8sutils.LargeResourceSizeLimitName)
+		}
+	}
+
+	log.ApplicationRestoreLog(restore).Infof("The size of application restore CR obtained %v bytes", rbCrSize)
+	rb.Status.TotalResourceCount = int64(len(rb.Status.Resources))
+	if rbCrSize > int(largeResourceSizeLimit) {
+		log.ApplicationRestoreLog(restore).Infof("Stripping all the resource info from restore cr as it is a large resource based restore")
```
prashanthpx, line 721: It would be good to add this also in the reason field for debugging purpose.

pkg/executor/nfs/nfsrestorevolcreate.go
```diff
@@ -288,10 +293,31 @@ func restoreVolResourcesAndApply(
...
+					rb.Status.Status = kdmpapi.ResourceBackupStatusInProgress
+					rb.Status.Reason = msg
+					rb.Status.ProgressPercentage = 0
+
+					_, err = kdmpschedops.Instance().UpdateResourceBackup(rb)
+					if err != nil {
+						errMsg := fmt.Sprintf("error updating ResourceBackup CR[%v/%v]: %v", rbCrNamespace, rbCrName, err)
+						return fmt.Errorf(errMsg)
+					}
+					//TODO: Should we also update the RE export CR here ? Or keep trying instead of return.
```
prashanthpx, line 315: Usually if we want to updated RE CR, it should be done in reconciler stage by reading the status and avoikd doing in the job.

pkg/apis/kdmp/v1alpha1/resourcebackup.go
```diff
@@ -34,6 +34,20 @@ const (
 	ResourceBackupStatusPartialSuccess ResourceBackupStatus = "PartialSuccess"
 )
 
+// ResourceApplyPhase defines a status of resource restoration phase.
+type ResourceApplyPhase string
```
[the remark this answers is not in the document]
lalat-das: Sure Will do

pkg/controllers/resourceexport/reconcile.go
```diff
@@ -172,6 +175,18 @@ func (c *Controller) process(ctx context.Context, in *kdmpapi.ResourceExport) (b
...
 				reason: "RestoreExport job in progress",
 			}
+			// When Resource-backup(RB) CR is In-PROGRESS, it is returned here till job is completed
+			// Lets update the Resource-Export(RE) CR with latests data here.
+			// Get the RB CR for writing, ignore if not created yet
+			rb, err := kdmp.Instance().GetResourceBackup(resourceExport.Name, resourceExport.Namespace)
+			if err != nil {
+				return true, c.updateStatus(resourceExport, updateData)
+			}
+			if rb.Status.ResourceApplyStage != kdmpapi.ResourceApplyPhase("") {
```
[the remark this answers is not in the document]
lalat-das: Yes prashanth it may happen that resourceApply's  resource-preparation stage wouldn't have started by the time RE reconciler kicked in and went for couple of rounds, in that case no point in updating these data.. But there is no side effect though updating these data uncondtionally., Trying to be more precise here..

pkg/executor/nfs/nfsrestoreresources.go
lalat-das: @prashanthpx This is the case of updating the progress bar related data in the RB CR, The thinking here was if it fails we need not fail the apply process as it will happen again after 5 mins. This is purely for progress bar % show up purpose.. eventual consistent can be adopted here.  Anyway if  some actual apply resource fails in between five minutes this data will show some stale count. So refering this values for strict debugging shouldn't be considered here.

PR #282, opened 2023-06-07
pb-3929: modify the error msg to be generic for kopia

lalat-das:
MountFailed related error need to be generic to be applicable for both S3 and NFS backuplocation scenario.

Closes #pb-3929


siva-portworx:
@lalat-das Can you please make sure this message is changed in the nfs driver as well?

lalat-das:
> @lalat-das Can you please make sure this message is changed in the nfs driver as well?

As discussed in case of nfs drivers will not be involved in case of S3, hence restarining from changing it.

PR #289, opened 2023-06-16
pb-3928: Fixed ivalid access of the volumesnapshot from the pvc spec with localsnapshot restore cleanup

siva-portworx:
```
pb-3928: Fixed ivalid access of the volumesnapshot from the pvc spec with localsnapshot restore cleanup
```
Closes # pb-3928

Tested parallel restore on the QA IBM setup.

### DIFF
--- a/pkg/controllers/dataexport/reconcile.go
+++ b/pkg/controllers/dataexport/reconcile.go
@@ -92,6 +92,7 @@ type updateDataExportDetail struct {
 	snapshotNamespace    string
 	removeFinalizer      bool
 	volumeSnapshot       string
+	resetLocalSnapshotRestore bool
 }
 
 func (c *Controller) sync(ctx context.Context, in *kdmpapi.DataExport) (bool, error) {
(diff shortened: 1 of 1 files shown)


pkg/controllers/dataexport/reconcile.go
```diff
@@ -136,6 +139,7 @@ func (c *Controller) sync(ctx context.Context, in *kdmpapi.DataExport) (bool, er
 		data := updateDataExportDetail{
 			status: kdmpapi.DataExportStatusFailed,
 			reason: msg,
+			localSnapshotRestore: dataExport.Status.LocalSnapshotRestore,
```
diptiranjanpx, line 142: Cant we add the updating of the flag `localSnapshotRestore` in updateStatus function and only update when it is absolutely needed. May be the default value can be false and only in one place where it has to become true , we can flip the value. Instead of updating in every dataexport updates.

siva-portworx: Modified the logic. Please check and approve.

PR #290, opened 2023-06-26
Don't check for BackupLocation with rsync driver

adityadani:


diptiranjanpx (approved):

adityadani:
@siva-portworx / @diptiranjanpx Can you check the 2nd commit as well? I needed some more changes. Planning to vendor this into stork for 23.7 release.

siva-portworx (approved):

PR #306, opened 2023-09-06
pb-4354: Added changes to mount the PVC parallel in kdmp job pod along with app pod.

siva-portworx:
```
 pb-4354: Added changes to mount the PVC parallel in kdmp job pod aling
    with app pod.

            - Remove the check for live and non-live backup
            - If the application pod is already running, the kdmp jpb pod
              will be scheduled on the same node.
            - If the application pod is not running, the kdmp job pod will
              scheduled in one of the node. If the app get scheduled during this time
              on different node for RWO access mode PVC, it will be stalled till the kdmp
              job pod is completed. Then the app pod will continue to run by mounting
              the pvc in it;s node, where it got scheduled.
```
Closes # pb-4354

This approach will not work for new access mode  ReadWriteOncePod. Will take this support later.
The above approach should work for remain access mode, except the case, whether the kdmp job got scheduled first for RWO and later the app pod was trying to mount the PVC. In that case, the app pod will staled till the kdmp backup pod completes.
Unit tested the below two cases. Rest will cover in functional test:
1) With RWO access mode, took backup when the PipelineRun pod was running. Made sure the kdmp job pod is scheduled on the same node as the app pod.
2) When none of the app pod was running started a backup. Once the backup kdmp job is scheduled, started the PipelineRun pod and checked that it is stalled as it started on different node and once the kdmp job pod was done, PipelineRun pod continue to run.


pkg/drivers/kopiabackup/kopiabackup.go
```diff
@@ -462,49 +467,31 @@ func buildJob(jobName string, jobOptions drivers.JobOpts) (*batchv1.Job, error)
...
-	count := len(pods)
 	for _, pod := range pods {
 		labels := pod.ObjectMeta.Labels
 		if _, ok := labels[drivers.DriverNameLabel]; ok {
-			count--
+			continue
+		}
+		if pod.Status.Phase == "Running" {
+			// get the nodeName, if the pods is in Running state, So that we can schedule
+			// kopis jon on the same node.
```
prashanthpx, line 479: Typo `kopis`

pkg/drivers/kopiabackup/kopiabackup.go
```diff
@@ -462,49 +467,31 @@ func buildJob(jobName string, jobOptions drivers.JobOpts) (*batchv1.Job, error)
...
+	var nodeName string
 	// filter out the pods that are create by us
-	count := len(pods)
 	for _, pod := range pods {
 		labels := pod.ObjectMeta.Labels
 		if _, ok := labels[drivers.DriverNameLabel]; ok {
-			count--
+			continue
+		}
+		if pod.Status.Phase == "Running" {
```
prashanthpx, line 477: Good to add comment that if app pod is in Pending state, then kopia job pod will let k8s schedule the pod.
We can handle Pending state later

prashanthpx (approved):

pkg/drivers/kopiabackup/kopiabackup.go
siva-portworx: Fixed.

pkg/drivers/kopiabackup/kopiabackup.go
siva-portworx: As discussed, we might not need special handling for Pending state.

PR #309, opened 2023-09-25
[PB-4396] Adding code to send Pod/job's description to stork in case of failure

shkumari-px:
Capture job's and pod's describe output in case of failure. 

Closes #PB-4396

Unit test result pending

### DIFF
--- a/pkg/controllers/dataexport/reconcile.go
+++ b/pkg/controllers/dataexport/reconcile.go
@@ -37,6 +37,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	k8shelper "k8s.io/component-helpers/storage/volume"
 )
@@ -508,6 +509,7 @@ func appendPodLogToStork(jobName string, namespace string) {
 	job, err := batch.Instance().GetJob(jobName, namespace)
 	if err != nil && !k8sErrors.IsNotFound(err) {
 		logrus.Infof("failed in getting job %v/%v with err: %v", namespace, jobName, err)
+		logrus.Infof("Reason for failure of the job: %s\n", job.Status.Conditions[0].Reason)
 	}
 	pods, err := core.Instance().GetPods(
 		job.Namespace,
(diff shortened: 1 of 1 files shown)


pkg/controllers/dataexport/reconcile.go
```diff
@@ -528,6 +530,25 @@ func appendPodLogToStork(jobName string, namespace string) {
 			logrus.Infof(podLog)
 			logrus.Infof("end of job-pod [%s]'s log...", pod.Name)
 		}
+
+		config, cerr := rest.InClusterConfig()
+		if cerr != nil {
+			panic(cerr.Error())
```
lalat-das, line 536: we don't need to panic the stork container here. 
Report error and pass the error back to caller if needed.

pkg/controllers/dataexport/reconcile.go
```diff
@@ -528,6 +530,25 @@ func appendPodLogToStork(jobName string, namespace string) {
...
+		}
+
+		clientset, cserr := kubernetes.NewForConfig(config)
+		if cserr != nil {
+			panic(cserr.Error())
+		}
+
+		Despod, Derr := clientset.CoreV1().Pods(pod.Namespace).Get(context.TODO(), pod.Name, metav1.GetOptions{})
```
lalat-das, line 544: We have already predefined APIs to get k8s client based on need. Please use what standard is followed already in code  
 core.Instance().GetPods()  or check in sched-ops  we should be having an equivalent API. 
 `	// GetPodByName returns pod for the given pod name and namespace
	GetPodByName(string, string) (*corev1.Pod, error)
`
PLease explore how we can use its return value

pkg/controllers/dataexport/reconcile.go
```diff
@@ -528,6 +530,25 @@ func appendPodLogToStork(jobName string, namespace string) {
...
+		if Derr != nil {
+			panic(Derr.Error())
+		}
+
+		if Despod.Status.Phase != "Running" {
```
lalat-das, line 549: variable name can be camel case and fully describabale.. Though it is not the way golang does like local varibales are named like a, b, c etc.. but we follow camelCase with higher verbosity in the name. PLease chnage DesPod, Derr accordingly.

pkg/controllers/dataexport/reconcile.go
```diff
@@ -528,6 +530,25 @@ func appendPodLogToStork(jobName string, namespace string) {
...
+			logrus.Infof("Pod Description by when Pod is not in running phase.....%s", Despod)
```
lalat-das, line 550: Would you like to reword the error msg ?

pkg/controllers/dataexport/reconcile.go
```diff
@@ -508,6 +509,7 @@ func appendPodLogToStork(jobName string, namespace string) {
 	job, err := batch.Instance().GetJob(jobName, namespace)
 	if err != nil && !k8sErrors.IsNotFound(err) {
 		logrus.Infof("failed in getting job %v/%v with err: %v", namespace, jobName, err)
+		logrus.Infof("Reason for failure of the job: %s\n", job.Status.Conditions[0].Reason)
```
lalat-das, line 512: Can we please add these detail  to above Info  message . 
PLease attach the unit test result for failure scenario by simulating a POD failure

shkumari-px:
It'll be tracked here - https://github.com/portworx/kdmp/pull/311

PR #315, opened 2023-10-11
pb-4579: Added check to ignore sseTypePath secret file as it is optional

siva-portworx:
```
pb-4579: Added check to ignore sseTypePath secret file as it is optional.
```

Closes #pb-4579

**Testing:**
On the QA cluster valid the deletion and maintenance job works for the backuplocaiton which did not had SseType field.

### DIFF
--- a/pkg/executor/common.go
+++ b/pkg/executor/common.go
@@ -349,9 +349,12 @@ func parseS3Creds() (*Repository, error) {
 
 	sseType, err := os.ReadFile(sseTypePath)
 	if err != nil {
-		errMsg := fmt.Sprintf("failed reading data from file %s : %s", sseTypePath, err)
-		logrus.Errorf("%v", errMsg)
-		return nil, fmt.Errorf(errMsg)
+		// sseTypePath field in the secret is optional, So ignoring if the file does not exists.
+		if !os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


pkg/executor/common.go
```diff
@@ -349,9 +349,12 @@ func parseS3Creds() (*Repository, error) {
...
 	if err != nil {
-		errMsg := fmt.Sprintf("failed reading data from file %s : %s", sseTypePath, err)
-		logrus.Errorf("%v", errMsg)
-		return nil, fmt.Errorf(errMsg)
+		// sseTypePath field in the secret is optional, So ignoring if the file does not exists.
+		if !os.IsNotExist(err) {
+			errMsg := fmt.Sprintf("failed reading data from file %s : %s", sseTypePath, err)
+			logrus.Errorf("%v", errMsg)
+			return nil, fmt.Errorf(errMsg)
+		}
```
prashanthpx, line 357: We also need to add general handling of error message if it fails for any other reason apart from IsNotExist()

prashanthpx (approved):

pkg/executor/common.go
siva-portworx: For other errors ( other than "file not found"), we will return error. in the case of error, byte array will be empty. That  means we will assign empty string as well.

PR #318, opened 2023-10-26
Added changes to avoid creationg of job cred secret when ssl is disabled

px-kesavan:
…led in bl

**What this PR does / why we need it**: Added changes to avoid creationg of job cred secret when ssl is disabled in backup location

Closes #


siva-portworx (approved):
I hope it is taken care during the cleanup as well.

PR #322, opened 2023-11-21
pb-4969: Fixed issue with backuplocation CR variable, which was causing the nfs backup issue.

siva-portworx:
```
   pb-4969: Fixed issue with backuplocation CR variable, which was causing
    the nfs backup issue.

            - backuplocation CR variable scope was lost inside the if
              condtion. Because of this the nfs parameters are not passed to kopia job.
            - The kopia writes the snapshot file directly  inside the job
              pod file system and files are lost once the pod is completed.
            - Since the nfs parameters are not passed, the nfs pvc was not
              getting mounted inside the nfs job pod.
```
Closes # pb-4969

Testing:
1) Tested both the backup and restore with NFS backuplocation.


lalat-das (approved):
Lgtm

prashanthpx (approved):

PR #324, opened 2023-11-27
Pb-4844: Made changes in kdmp for it to enable debug level log messages

kgarg-px:
This PR introduces changes to include the debug level of logs for all kopia related operations. Needed, so it can be easier to debug any messages in the future when required.
Closes #

Attaching screenshots for reference 
Screenshot 2023-11-27 at 4 05 47 PM
Screenshot 2023-11-27 at 4 12 44 PM
Screenshot 2023-11-27 at 4 15 43 PM
Screenshot 2023-11-27 at 10 50 26 AM

### DIFF
--- a/pkg/drivers/kopiabackup/kopiabackup.go
+++ b/pkg/drivers/kopiabackup/kopiabackup.go
@@ -287,6 +287,9 @@ func jobFor(
 		"/data",
 	}, " ")
 
+	// Read the config map from the job option and then add the debug level log if required
+	cmd = utils.CheckAndAddKopiaDebugModeEnabled(cmd, jobOption)
+
 	if jobOption.Compression != "" {
 		splitCmd := strings.Split(cmd, " ")
 		splitCmd = append(splitCmd, "--compression", jobOption.Compression)
--- a/pkg/drivers/kopiarestore/kopiarestore.go
+++ b/pkg/drivers/kopiarestore/kopiarestore.go
@@ -206,6 +206,8 @@ func jobFor(
 		vb.Status.SnapshotID,
 	}, " ")
 
+	cmd = utils.CheckAndAddKopiaDebugModeEnabled(cmd, jobOption)
+
 	kopiaExecutorImage, imageRegistrySecret, err := utils.GetExecutorImageAndSecret(drivers.KopiaExecutorImage,
 		jobOption.KopiaImageExecutorSource,
 		jobOption.KopiaImageExecutorSourceNs,
@@ -369,7 +371,7 @@ func roleFor() *rbacv1.Role {
 			},
 			{
 				APIGroups: []string{"kdmp.portworx.com"},
-				Resources: []string{"volumebackups"},
+				Resources: []string{"volumebackups", "dataexports"},
 				Verbs:     []string{rbacv1.VerbAll},
 			},
 		},
--- a/pkg/executor/kopia/kopiabackup.go
+++ b/pkg/executor/kopia/kopiabackup.go
@@ -69,7 +69,6 @@ func newBackupCommand() *cobra.Command {
 	backupCommand.Flags().StringVar(&sourcePathGlob, "source-path-glob", "", "The regexp should match only one path that will be used for backup")
 	backupCommand.Flags().StringVar(&compression, "compression", "", "Compression type to be used")
 	backupCommand.Flags().StringVar(&excludeFileList, "exclude-file-list", "", " list of dir names that need to be exclude in the kopia snapshot")
-
 	return backupCommand
 }
 
@@ -249,6 +248,9 @@ func runKopiaCreateRepo(repository *executor.Repository) error {
 	if err != nil {
 		return err
 	}
+	// Now check if debug log level is to be added in the command which got prepared above.
+	// check if debug level exists
+	repoCreateCmd = addLogLevelDebugToCommand(repoCreateCmd, logLevelDebug)
 	// NFS doesn't need any special treatment for repo create command
 	// hence no case exist for it.
 	switch repository.Type {
@@ -323,6 +325,8 @@ func runKopiaBackup(repository *executor.Repository, sourcePath string) error {
 	if err != nil {
 		return err
 	}
+	// Check and add debug level logs for kopia backup command
+	backupCmd = addLogLevelDebugToCommand(backupCmd, logLevelDebug)
 	// This is needed to handle case where after kopia repo create was successful and
 	// the pod got terminated. Now user triggers another backup, so we need to pass
 	// credentials for "snapshot create".
@@ -386,6 +390,8 @@ func runKopiaRepositoryConnect(repository *executor.Repository) error {
 	if err != nil {
 		return err
 	}
+	// Check and add debug level logs for kopia connect command
+	connectCmd = addLogLevelDebugToCommand(connectCmd, logLevelDebug)
 
 	switch repository.Type {
 	case storkv1.BackupLocationS3:
@@ -430,6 +436,7 @@ func setGlobalPolicy() error {
 	if err != nil {
 		return err
 	}
+	policyCmd = addLogLevelDebugToCommand(policyCmd, logLevelDebug)
 	// As we don't want kopia maintenance to kick in and trigger global policy on default values
 	// for the repository, setting them to very high values
 	policyCmd = addPolicySetting(policyCmd)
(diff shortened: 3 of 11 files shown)


pkg/executor/kopia/kopiabackup.go
```diff
@@ -261,7 +263,7 @@ func runKopiaCreateRepo(repository *executor.Repository) error {
 	}
 
 	initExecutor := kopia.NewCreateExecutor(repoCreateCmd)
-	if err := initExecutor.Run(); err != nil {
+	if err := initExecutor.Run(logLevel); err != nil {
```
vsundarraj-px, line 266: I think too much changes have happened because the basic signature has been changed. The best way is to use use internal options than as function argument. You can follow how other options have been used without changing Run() signature and follow the same. 

This applies to all places where existing function signature is changed to add loglevel.

nit:  in general log-levels has multiple levels. Info/error/trace/debug. I see log-level means debug only. .. You should use specific name if there are no multiple values it can take. if debug is what you mean specifically, you should change it to enable-debug-logging or log-level-debug

vsundarraj-px (changes requested):

pkg/executor/kopia/kopiabackup.go
prashanthpx: +1 for the above

pkg/drivers/kopiabackup/kopiabackup.go
```diff
@@ -258,6 +260,20 @@ func (d Driver) validate(o drivers.JobOpts) error {
 	return nil
 }
 
+func checkDataExportAnnotation(jobOption drivers.JobOpts) bool {
```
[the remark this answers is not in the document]
kgarg-px: made changes accordingly

pkg/drivers/kopiarestore/kopiarestore.go
```diff
@@ -206,6 +208,12 @@ func jobFor(
 		vb.Status.SnapshotID,
 	}, " ")
 
+	if checkDataExportAnnotation(jobOption) {
```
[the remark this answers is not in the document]
kgarg-px: Made changes

pkg/drivers/kopiarestore/kopiarestore.go
```diff
@@ -369,9 +377,23 @@ func roleFor() *rbacv1.Role {
...
 				APIGroups: []string{"kdmp.portworx.com"},
-				Resources: []string{"volumebackups"},
+				Resources: []string{"volumebackups", "dataexports"},
 				Verbs:     []string{rbacv1.VerbAll},
 			},
 		},
 	}
 }
+
+func checkDataExportAnnotation(jobOption drivers.JobOpts) bool {
```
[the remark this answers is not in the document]
kgarg-px: Moved the function to pkg/drivers/utils/utils.go

pkg/executor/kopia/kopiabackup.go
kgarg-px: Made the suggested changes and tested as well with new changes

vsundarraj-px (changes requested):

siva-portworx:
@kgarg-px Can you add the new debug message that we are getting, when you enabled debug option for kopia? In the screenshot i do not see any new messsage from kopia tool code. All the messages that i see are from kopia-executor code.

kgarg-px:
> @kgarg-px Can you add the new debug message that we are getting, when you enabled debug option for kopia? In the screenshot i do not see any new messsage from kopia tool code. All the messages that i see are from kopia-executor code.

@siva-portworx 
The only new messages from kopia tool are visible only in the case of `backup` and `maintenance` at times.

pkg/executor/kopia/kopiabackup.go
```diff
@@ -677,3 +699,17 @@ func addPolicySetting(policyCmd *kopia.Command) *kopia.Command {
 
 	return policyCmd
 }
+
+func isKopiaDebugModeEnabled(kopiaCommand *kopia.Command, logLevelDebug string) *kopia.Command {
```
vsundarraj-px, line 703: You should remove this function and add this check in addLogLevelDebugtoCommand() 
isKopiaDebugModeEnabled() sounds like check function and should not set any thing in this. 
But as suggested this function can be removed.

pkg/executor/kopia/kopiadelete.go
```diff
@@ -133,6 +135,9 @@ func runKopiaDelete(repository *executor.Repository, snapshotID string) error {
 		logrus.Errorf("%s %v", fn, errMsg)
 		return fmt.Errorf(errMsg)
 	}
+	// Check and add debug level logs for kopia delete command
+	deleteCmd = isKopiaDebugModeEnabled(deleteCmd, logLevelDebug)
```
vsundarraj-px, line 139: As suggested above replace this with addLogLevelDebugToCommand() after moving the check code inside this function

pkg/executor/kopia/maintenance.go
```diff
@@ -281,6 +283,8 @@ func runKopiaQuickMaintenanceExecute(repository *executor.Repository) error {
 		return fmt.Errorf(errMsg)
 	}
 
+	// Check and add debug log level for kopia maintenance command
+	maintenanceRunCmd = isKopiaDebugModeEnabled(maintenanceRunCmd, logLevelDebug)
```
vsundarraj-px, line 287: Same here..and every place where isKopiaDebugEnabled() is called.

pkg/executor/kopia/restore.go
```diff
@@ -108,6 +110,9 @@ func runKopiaRestore(repository *executor.Repository, targetPath, snapshotID str
 		return err
 	}
 
+	// Check and add debug level logs for kopia restore command
+	restoreCmd = isKopiaDebugModeEnabled(restoreCmd, logLevelDebug)
+
```
vsundarraj-px, line 115: remove extra space

pkg/kopia/command.go
```diff
@@ -191,6 +191,7 @@ func (c *Command) BackupCmd() *exec.Cmd {
 		configFile,
 		"--json",
 	}
+
```
vsundarraj-px, line 194: Remove all extra lines

vsundarraj-px (changes requested):

vsundarraj-px (approved):
LGTM

pkg/executor/kopia/kopiabackup.go
```diff
@@ -249,6 +251,9 @@ func runKopiaCreateRepo(repository *executor.Repository) error {
 	if err != nil {
 		return err
 	}
+	// Now check if debug log level is to be added in the command which got prepared above.
+	// check if debug level exists
+	repoCreateCmd = addLogLevelDebugToCommand(repoCreateCmd, logLevelDebug)
```
prashanthpx, line 256: Ideally we should move this adding to args in command.go file as this is commong to all flow path. We can set to enable debug mode in `type Command` struct and use that to fill details in all places.

pkg/drivers/kopiabackup/kopiabackup.go
```diff
@@ -507,7 +509,7 @@ func roleFor(live bool) *rbacv1.Role {
 		Rules: []rbacv1.PolicyRule{
 			{
 				APIGroups: []string{"kdmp.portworx.com"},
-				Resources: []string{"volumebackups"},
+				Resources: []string{"volumebackups", "dataexports"},
```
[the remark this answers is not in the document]
kgarg-px: @prashanthpx The reason why we are using DE CR here is because, we are setting certain annotations in the DE CR while reading the key-value pair from the `kdmp-config` CM, if a particular key-value pair is present, then we set the annotation in the DE CR, and in kdmp, we are reading the DE CR and seeing if the particular annotation is present, upon which we enable the debug mode in kopia

pkg/executor/kopia/kopiabackup.go
prashanthpx: Fine for now but flow is for few things which are very specific to command we add it and common things goes to Command struct.

pkg/drivers/kopiabackup/kopiabackup.go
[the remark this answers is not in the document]
prashanthpx: Looking at the stork PR, my understanding on the flow is 
1. You read the CM in stork
2. Pass the Debug mode in the DE CR

Indeed we need not do this way as reconciler code in kdmp does read `kdmp-config` value (for eg check `KopiaResourceRequirements`) where we can set kdmp job resoruces on the fly.
We should be following the same with which, we won't need any stork change except vendoring the code to stork

prashanthpx (approved):
LGTM. Before merging can you pls remove the vendor code of strok as it is not needed.

PR #336, opened 2024-01-28
pb-5288: Added changes to read the istio disable config from px-backup-config for delete and maintenance apis.

siva-portworx:
```
 pb-5288: Added changes to read the istio disable config from px-backup-config for delete and maintenance apis.
```
Closes #pb-5288

Will add the test result with px-backup PR.

### DIFF
--- a/pkg/drivers/utils/utils.go
+++ b/pkg/drivers/utils/utils.go
@@ -878,10 +878,10 @@ func IsJobPodMountFailed(job *batchv1.Job, namespace string) bool {
 	return false
 }
 
-func GetDisableIstioConfig() bool {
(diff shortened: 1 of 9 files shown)


pkg/drivers/utils/utils.go
```diff
@@ -27,7 +27,14 @@ import (
 )
 
 const (
-	defaultPXNamespace = "kube-system"
+	// BackupConfigMap holds any config needs to be provided to job
+	// from the caller. Eg: executor image name, secret, etc..
+	BackupConfigMap = "px-backup-config"
```
prashanthpx, line 32: Thinking shouldn't we use the existing `kdmp-config`  map already present which is used for all jobs parameterization rather than puhsing the istio only label to `px-backup-config` as this is not really used by px-backup pod right?

prashanthpx (approved):

PR #346, opened 2024-03-11
added log to identify if it's a resource only backup

shkumari-px:
**What this PR does / why we need it**: It adds logs to identify if it's a resource only backup in case of s3 bl

Closes # [PB-5823](https://portworx.atlassian.net/browse/PB-5823)

**RCA**-
If we take resource only backup (without any pvc), then we don't create a generic backup dir on the backup location
**TestMaintenanceJobForResourceOnlyBackup** checks for the log in the maintenance pod log to verify if it's a resource only backup
If we don't find any generic backup dir, then we throw this warning "No directory exists, verify if it is a resource only backup"
This log statement is covered in nfs backup flow but not in s3 bl scenario
**ss of quick maintenance pod log**
Screenshot 2024-03-11 at 10 13 32 AM
Screenshot 2024-03-11 at 11 21 38 AM

[PB-5823]: https://portworx.atlassian.net/browse/PB-5823?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/executor/kopia/maintenance.go
+++ b/pkg/executor/kopia/maintenance.go
@@ -171,6 +171,10 @@ func runMaintenance(maintenanceType string) error {
 		repo.Name = genericBackupDir + "/"
 		bucket = blob.PrefixedBucket(bucket, repo.Name)
 		repoList, err = getRepoList(bucket)
+		if len(repoList) == 0 {
+			logrus.Warnf("Provider is non-nfs, No directory %v exists, verify if it is a resource only backup", repo.Name)
+			return nil
+		}
 		if err != nil {
 			logrus.Errorf("getting repo list failed for bucket [%v]: %v", repo.Path, err)
 			return err


pkg/executor/kopia/maintenance.go
```diff
@@ -171,6 +171,10 @@ func runMaintenance(maintenanceType string) error {
 		repo.Name = genericBackupDir + "/"
 		bucket = blob.PrefixedBucket(bucket, repo.Name)
 		repoList, err = getRepoList(bucket)
+		if len(repoList) == 0 {
+			logrus.Warnf("Provider is non-nfs, No directory %v exists, verify if it is a resource only backup", repo.Name)
```
siva-portworx, line 175: Here repo name will be "genericBackupDir/" It is better to add the backup location CR name in the debug statement.

siva-portworx (approved):

PR #347, opened 2024-03-17
PWX-36486: Rsync job name to be restricted within max length limit.

diptiranjanpx:
Signed-Off-By: Diptiranjan

**Unit test result**:
```
➜  go test -v -tags unittest -coverprofile=profile.out -covermode=atomic  github.com/portworx/kdmp/pkg/drivers/rsync

=== RUN   TestToJobName
--- PASS: TestToJobName (0.00s)
PASS
coverage: 8.3% of statements
ok      github.com/portworx/kdmp/pkg/drivers/rsync      1.776s  coverage: 8.3% of statements
```

The fix is for making rsync job name within job name limit.
Also added dataexport UID to create new job everytime if dataexport spec gets reapplied with same name.

Closes # PWX_36486

Test results have been updated in PWX-36486


pp511 (approved):
looks ok

PR #355, opened 2024-03-29
PB-6205: Support for NFS related partial backup and restore of those partial backups

kgarg-px:
This PR introduces two changes where when the objects are uploaded to NFS location. Now, we check if any of the PV and PVCs have failed and skip those while uploading the `resources.json` file. 
During restore from NFS in case if it is partial backup, we skip the PVCs which were either in `Failed` or `Skipped` state
Closes #

Restore is successful when the backup was in partial backup state
Screenshot 2024-03-29 at 7 21 58 PM

PVC which was in failed state during backup is skipped during restore and the restore status is as successful:
Screenshot 2024-03-29 at 7 21 48 PM

Logs of nfs executor during restore:
Screenshot 2024-03-29 at 7 20 58 PM
Screenshot 2024-03-29 at 7 21 11 PM

As seen below, in the resources.json file, the PVC which was in failed state is not uploaded in the file in the BL
Screenshot 2024-03-29 at 7 13 06 PM

### DIFF
--- a/pkg/executor/nfs/nfsbkpresources.go
+++ b/pkg/executor/nfs/nfsbkpresources.go
@@ -35,6 +35,7 @@ import (
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


prashanthpx (approved):
LGTM. Can you please split up vendor into a separate PR. For this PR lets only have the feature changes.

kgarg-px:
> LGTM. Can you please split up vendor into a separate PR. For this PR lets only have the feature changes.

@prashanthpx Made the PR for just the feature changes.

PR #360, opened 2024-04-22
PB-6687 :: Placing PX secret in nfsrestore job spec 

vikasit12:
This PR gets secret env from stork deploy used in nfs restore of secured portworx volumes. The access key retrieved is passed as env variable in nfs restore job to complete nfs restore

Closes #PB-6687

image
image

**Test Cases combination performed**:
Backup and restore
1. Portworx (Across Cluster)
2. Portworx + CSI + OFFLOAD ( Across Cluster )
3. Portworx + KDMP ( Across Cluster )
4. Portworx + CSI ( Same Cluster ) ( Retain ) ( Deleted App before Restore )
5. BACKUP_TYPE = Generic

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 /bin
 profile.out
 coverage.txt
+.vscode/
+.idea/
--- a/pkg/executor/nfs/nfsrestorevolcreate.go
+++ b/pkg/executor/nfs/nfsrestorevolcreate.go
@@ -135,6 +135,11 @@ func restoreVolResourcesAndApply(
 			if volumeBackup.Namespace != namespace {
 				continue
 			}
+
+			// TODO: Skip portworx volumes in case of nfsRestore PB:6687
+			if driverName == "pxd" {
+				continue
+			}
 			// If a list of resources was specified during restore check if
 			// this PVC was included
 			info.Name = volumeBackup.PersistentVolumeClaim


lalat-das (approved):

vsundarraj-px (commented):
This review is generated on the latest stork. Can you please rebase with master and then send PR. There are stale changes. 

Also if there is vendor changes. Best to send it as seperate PR to have only the necessary changes show up in commit.

pkg/drivers/nfsrestore/nfsrestore.go
```diff
@@ -317,6 +322,18 @@ func jobForRestoreResource(
 	if len(imageRegistrySecret) != 0 {
 		job.Spec.Template.Spec.ImagePullSecrets = utils.ToImagePullSecret(utils.GetImageSecretName(jobOption.RestoreExportName))
 	}
+	if len(pxSecretIssuer) != 0{
+		job.Spec.Template.Spec.Containers[0].Env = append(job.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{
+			Name:  k8sutils.PxSecretIssuerEnvName,
+			Value: pxSecretIssuer,
+		})
+	}
+	if len(pxSecretKey) != 0{
```
vsundarraj-px, line 331: By setting the secretKey as ENV would make this visible to every one who has even pod view. If I understand from stork. The stork deployment only references the k8s secret reference name but the secretKey is embedded in the k8s secret. But here we are passing as job pod env. I am not sure if this becomes a security vulnerability.

vikasit12: > By setting the secretKey as ENV would make this visible to every one who has even pod view. If I understand from stork. The stork deployment only references the k8s secret reference name but the secretKey is embedded in the k8s secret. But here we are passing as job pod env. I am not sure if this becomes a security vulnerability.

This is a nfs restore job so once completed it gets deleted. Also stork can use that in such a manner because it resides in same namespace as of portworx but nfs restore will start in some other ns where that secret will not be present hence passing that as ENV.

Also anyone having job pod view right now can view these access keys from stork too as in pod it is an ENV variable only

siva-portworx: @vikasit12 As discussed, let us please create the secret in the stork code ( Resourceexport reconciler code) and pass only the secret name and namespace similar to other secret like cred/cert/image registry secret.

siva-portworx (approved):

PR #371, opened 2024-05-07
pb-6681: Add support for PSA in kdmp job

lalat-das:
- Added check and collected psa info from the namespace
- if psa enabled then extracted the uid and gid used by the POD. here the pod is choosen based on whichever PVC is used by that pod
- applied those uid and GID to all relevant job pod spec
- this is done only if the PSA mode enforced with "restricted" value
- for baseline and privilege mode no restriction on UID/GID and default setting of SElinux and secomp is adopted in the POD spec

**What this PR does / why we need it**: this adds the PSA feature to px-backup

Closes # pb-6681

Currently Unit testing, this is a PR to get hold of the code review earlier.
Known Issue : restore path doesn't have a deployment to capture UID and GID, we will use the backupInfo artifacts in mongo to provide that. we should have captured it during backup time.

### DIFF
--- a/pkg/drivers/utils/utils.go
+++ b/pkg/drivers/utils/utils.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -67,6 +68,10 @@ const (
 	IstioInjectLabel  = "sidecar.istio.io/inject"
 	// ProcessVMResourceSuccessMsg - vm resources processed successfully
 	ProcessVMResourceSuccessMsg = "vm resources processed successfully"
+	// podSecurityStandardEnforceMode - which PS standard to enforce while deploying the k8s workload Object
+	podSecurityStandardEnforceMode = "pod-security.kubernetes.io/enforce"
+	//undefined UID or GID
+	UndefinedId = int64(-1)
 )
 
 var (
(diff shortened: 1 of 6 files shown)


pkg/drivers/utils/utils.go
```diff
@@ -966,3 +971,159 @@ func GetShortUID(uid string) string {
...
+	gid := UndefinedId
+	// Check if the namespace has the PSA labels set
+	psaIsEnforced, psaLabelMap, err := GetPsaDetail(namespace)
+	if err != nil {
+		logrus.Errorf("%s: %v", fn, err)
+		return uid, gid, err
+	}
+	if !psaIsEnforced {
+		errMsg := fmt.Sprintf("PSA is not enforced in namespace [%s], no need to enforce any uid/gid in job for backing up pvc [%s]", namespace, pvcName)
+		logrus.Debugf("%s: %v", fn, errMsg)
```
px-kesavan, line 1106: Here its a debug , but in the caller we are logging error if PSA is not enforced

pkg/drivers/utils/utils.go
```diff
@@ -966,3 +971,159 @@ func GetShortUID(uid string) string {
...
+				psaIsEnforced = true
+			}
+		}
+	}
+	return psaIsEnforced, psaLabelMap, nil
+}
+
+// get Pod from PVC details
+func GetPodFromPVC(pvcName, namespace string) (*corev1.Pod, error) {
+	fn := "GetPodFromPVCName"
```
px-kesavan, line 1008: Nit, change it to same as function name

lalat-das:
Created a new PR and addressed the review comment given by kesavan in that new PR. That PR is created since we need to vendor a different stork version from private to master. And a lot other changes to be addressed hence made a separate branch and a PR eventually.

PR #372, opened 2024-05-09
PB-6687 :: Skip Portworx Volume in Case NFS Restore

vikasit12:

Closes #PB-6687


pkg/executor/nfs/nfsrestorevolcreate.go
```diff
@@ -135,6 +135,12 @@ func restoreVolResourcesAndApply(
 			if volumeBackup.Namespace != namespace {
 				continue
 			}
+
+			// Skip portworx volumes in case of nfsRestore PB-6687
+			if driverName == "pxd" {
+				continue
```
siva-portworx, line 141: Can we add a debug statement with  volume name when we are skipping?

siva-portworx (approved):

PR #376, opened 2024-05-14
PB-6853: travis changes to push docker images for private branches

dbinnal-px:
Travis changes to push the docker images for private branches with branch name as tag.

Closes #

Screenshot 2024-05-14 at 11 24 31 AM
Screenshot 2024-05-14 at 11 23 45 AM

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ script:
     elif [ "${TRAVIS_BRANCH}" == "master" ]; then
       export DOCKER_IMAGE_TAG=master-latest
     else
-      export RELEASE_VER=`git rev-parse --short HEAD`
+      export DOCKER_IMAGE_TAG=${TRAVIS_BRANCH}
     fi
     make all &&
     if [ "${TRAVIS_PULL_REQUEST}" == "false" ]; then


.travis.yml
```diff
@@ -13,7 +13,7 @@ script:
     elif [ "${TRAVIS_BRANCH}" == "master" ]; then
       export DOCKER_IMAGE_TAG=master-latest
     else
-      export RELEASE_VER=`git rev-parse --short HEAD`
+      export DOCKER_IMAGE_TAG=${TRAVIS_BRANCH}
```
shkumari-px, line 16: We can do the same for px-backup as well @dbinnal-px

dbinnal-px: Good point, Will check with @prashanthpx

prashanthpx: Thinking back, the reason they had git sha was a docker image build was always available for every commit we make and is consumable. When this changes to branch, then there are possibility that more than one people working on the same branch will have images overwritten. I think then lets keep the change as is before.

dbinnal-px: Closing the PR

PR #377, opened 2024-05-16
Pb-6681 : added support for PSA for kdmp jobs

lalat-das:
**What this PR does / why we need it**: This enables the KDMP job pods to be added with SecurityContext setting required for it to run in a restricted PSA environment.

Closes # pb-6681

Cluster Wide PSA setting related code is not added yet.
Will update unit test result before EOD tomorrow

### DIFF
--- a/pkg/drivers/kopiabackup/kopiabackup.go
+++ b/pkg/drivers/kopiabackup/kopiabackup.go
@@ -318,6 +318,7 @@ func jobFor(
 		logrus.Errorf("failed to get the toleration details: %v", err)
 		return nil, fmt.Errorf("failed to get the toleration details for job [%s/%s]", jobOption.Namespace, jobName)
 	}
+
 	job := &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      jobName,
@@ -383,6 +384,13 @@ func jobFor(
 				},
 			},
 		},
+	}
+	// Add security Context only if the PSA is enabled.
+	if jobOption.PsaIsEnabled == "true" {
+		job, err = utils.AddSecurityContextToJob(job, jobOption.PodUserId, jobOption.PodGroupId)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	if len(nodeName) != 0 {
--- a/pkg/drivers/kopiarestore/kopiarestore.go
+++ b/pkg/drivers/kopiarestore/kopiarestore.go
@@ -219,6 +219,7 @@ func jobFor(
 		logrus.Errorf("failed to get the toleration details: %v", err)
 		return nil, fmt.Errorf("failed to get the toleration details for job [%s/%s]", jobOption.Namespace, jobName)
 	}
+
 	job := &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      jobName,
@@ -285,7 +286,13 @@ func jobFor(
 			},
 		},
 	}
-
+	// Add security Context only if the PSA is enabled.
+	if jobOption.PsaIsEnabled == "true" {
+		job, err = utils.AddSecurityContextToJob(job, jobOption.PodUserId, jobOption.PodGroupId)
+		if err != nil {
+			return nil, err
+		}
+	}
 	// Add the image secret in job spec only if it is present in the stork deployment.
 	if len(imageRegistrySecret) != 0 {
 		job.Spec.Template.Spec.ImagePullSecrets = utils.ToImagePullSecret(utils.GetImageSecretName(jobName))
--- a/pkg/drivers/nfsbackup/nfsbackup.go
+++ b/pkg/drivers/nfsbackup/nfsbackup.go
@@ -199,7 +199,6 @@ func jobForBackupResource(
 	}, " ")
 
 	labels := addJobLabels(jobOption)
-
 	nfsExecutorImage, imageRegistrySecret, err := utils.GetExecutorImageAndSecret(drivers.NfsExecutorImage,
 		jobOption.NfsImageExecutorSource,
 		jobOption.NfsImageExecutorSourceNs,
@@ -216,6 +215,7 @@ func jobForBackupResource(
 		logrus.Errorf("failed to get the toleration details: %v", err)
 		return nil, fmt.Errorf("failed to get the toleration details for job [%s/%s]", jobOption.Namespace, jobOption.RestoreExportName)
 	}
+
 	job := &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      jobOption.RestoreExportName,
@@ -270,6 +270,18 @@ func jobForBackupResource(
 			},
 		},
 	}
+
+	// Add security Context only if the PSA is enabled.
+	if jobOption.PsaIsEnabled == "true" {
+		// The Job is intended to backup resources to  NFS backuplocation
+		// and it doesn't need a specific JOB uid/gid since it will be sqaushed at NFS server
+		// hence used a global hardcoded UID/GID.
+		job, err = utils.AddSecurityContextToJob(job, utils.KdmpJobUid, utils.KdmpJobGid)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	// Add the image secret in job spec only if it is present in the stork deployment.
 	if len(imageRegistrySecret) != 0 {
 		job.Spec.Template.Spec.ImagePullSecrets = utils.ToImagePullSecret(utils.GetImageSecretName(jobOption.RestoreExportName))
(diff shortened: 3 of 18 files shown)


pkg/drivers/kopiabackup/kopiabackup.go
```diff
@@ -359,6 +371,18 @@ func jobFor(
 									ReadOnly:  true,
 								},
 							},
+							SecurityContext: &corev1.SecurityContext{
```
siva-portworx, line 374: AddSecurityContextToJob utility function already adding these fields. Do we need this below code?

pkg/drivers/kopiarestore/kopiarestore.go
```diff
@@ -219,6 +220,18 @@ func jobFor(
 		logrus.Errorf("failed to get the toleration details: %v", err)
 		return nil, fmt.Errorf("failed to get the toleration details for job [%s/%s]", jobOption.Namespace, jobName)
 	}
+
+	// get dataexport CR from itds name and namespace
```
siva-portworx, line 224: typo itds

pkg/drivers/nfsbackup/nfsbackup.go
```diff
@@ -270,6 +283,22 @@ func jobForBackupResource(
...
+	re, err := kdmp.Instance().GetResourceExport(jobOption.RestoreExportName, jobOption.Namespace)
+	if err != nil {
+		errMsg := fmt.Sprintf("failed to get resourceExport CR %s/%s: %v", jobOption.Namespace,jobOption.RestoreExportName, err)
+		logrus.Errorf("%v", errMsg)
+		return nil, fmt.Errorf(errMsg)
+	}
+	psaIsEnabled := re.Annotations[utils.PsaEnabledKey]
+	// Add security Context only if the PSA is enabled.
+	if psaIsEnabled == "true" {
+		job, err = utils.AddSecurityContextToJob(job, "", "")
```
vsundarraj-px, line 296: Comment would be better to understand why we dont need userID and Gid here.
And its also better if you send kdmpJobUid since its as static ID globally declared.

pkg/drivers/utils/utils.go
```diff
@@ -966,3 +972,45 @@ func GetShortUID(uid string) string {
 	}
 	return uid[:8]
 }
+
+// Add container security Context to job pod if the PSA is enabled.
+func AddSecurityContextToJob(job *batchv1.Job, podUserId, podGroupId string) (*batchv1.Job, error) {
+	if job != nil {
+		if podUserId != "" {
```
vsundarraj-px, line 979: This whole if chunk is not required if you pass kdmpJobUid by the caller itself.

pkg/drivers/utils/utils.go
```diff
@@ -966,3 +972,45 @@ func GetShortUID(uid string) string {
...
+			if err != nil {
+				logrus.Errorf("failed to convert the UID to int: %v", err)
+				return nil, fmt.Errorf("failed to convert the UID to int: %v", err)
+			}
+			job.Spec.Template.Spec.SecurityContext.RunAsUser = &uid
+		} else {
+			// we need to add a default user id if the user id is not provided
+			job.Spec.Template.Spec.SecurityContext.RunAsUser = pointer.Int64(kdmpJobUid)
+		}
+		if podGroupId != "" {
```
vsundarraj-px, line 990: Same here

pkg/drivers/kopiarestore/kopiarestore.go
```diff
@@ -260,6 +273,18 @@ func jobFor(
 									ReadOnly:  true,
 								},
 							},
+							SecurityContext: &corev1.SecurityContext{
```
vsundarraj-px, line 276: why are we setting this by default for every job spinned? Arent they only needed if psaIsEnabled == true?

pkg/drivers/kopiabackup/kopiabackup.go
```diff
@@ -318,6 +319,17 @@ func jobFor(
 		logrus.Errorf("failed to get the toleration details: %v", err)
 		return nil, fmt.Errorf("failed to get the toleration details for job [%s/%s]", jobOption.Namespace, jobName)
 	}
+	// get dataexport CR and read psa specific labels.
+	de, err := kdmpops.Instance().GetDataExport(context.Background(), jobOption.DataExportName, jobOption.Namespace)
```
prashanthpx, line 323: Rather than reading in every jobFor() function, you can read it once in the reconciler and pass it as jobOption parameters.

pkg/drivers/nfsbackup/nfsbackup.go
```diff
@@ -253,6 +254,18 @@ func jobForBackupResource(
 									ReadOnly:  true,
 								},
 							},
+							SecurityContext: &corev1.SecurityContext{
+								RunAsNonRoot:             pointer.Bool(true),
```
prashanthpx, line 258: We would need to remove this and piggyback on `AddSecurityContextToJob()` function

pkg/drivers/kopiabackup/kopiabackup.go
lalat-das: Thanks I need to remove this.. It was a remnant of approach -01

pkg/drivers/kopiabackup/kopiabackup.go
lalat-das: Make sense thanks for the catch.

pkg/drivers/utils/utils.go
lalat-das: done

pkg/drivers/utils/utils.go
lalat-das: Done. Kept the if check for nil string check but removed the else part though which is not needed if we pass the actual UID

pkg/drivers/nfsbackup/nfsbackup.go
lalat-das: Done

pkg/drivers/kopiarestore/kopiarestore.go
lalat-das: Done

pkg/drivers/nfsbackup/nfsbackup.go
lalat-das: Yeah removed it..it was stale code from approach -01 missed to be deleted.

pkg/drivers/utils/utils.go
```diff
@@ -966,3 +972,42 @@ func GetShortUID(uid string) string {
...
+		job.Spec.Template.Spec.Containers[0].SecurityContext.SeccompProfile = &corev1.SeccompProfile{
+			Type: "RuntimeDefault",
+		}
+		job.Spec.Template.Spec.Containers[0].SecurityContext.Capabilities = &corev1.Capabilities{
+			Drop: []corev1.Capability{
+				"ALL",
+			},
+		}
+	} else {
+		return job, fmt.Errorf("recieved a nil job object to add security context")
```
siva-portworx, line 1010: recieved : typo

pkg/drivers/utils/utils.go
```diff
@@ -966,3 +972,42 @@ func GetShortUID(uid string) string {
...
+		job.Spec.Template.Spec.Containers[0].SecurityContext.AllowPrivilegeEscalation = pointer.Bool(false)
+		job.Spec.Template.Spec.Containers[0].SecurityContext.SeccompProfile = &corev1.SeccompProfile{
+			Type: "RuntimeDefault",
+		}
+		job.Spec.Template.Spec.Containers[0].SecurityContext.Capabilities = &corev1.Capabilities{
+			Drop: []corev1.Capability{
+				"ALL",
+			},
+		}
+	} else {
```
vsundarraj-px, line 1009: We could avoid if and else and indentation if the code is 

if job == nil {
    return fmt.Errorf(...)
}
Rest of the code here..

pkg/drivers/utils/utils.go
```diff
@@ -966,3 +972,42 @@ func GetShortUID(uid string) string {
 	}
 	return uid[:8]
 }
+
+// Add container security Context to job pod if the PSA is enabled.
+// if static uids like kdmpJobUid or kdmpJobGid is used that means
+// these are dummy UIDs used for backing up resources to backuplocation
+// which doesn't need specific UID specific permission.
+func AddSecurityContextToJob(job *batchv1.Job, podUserId, podGroupId string) (*batchv1.Job, error) {
```
vsundarraj-px, line 980: Why is the job needs to be returned. You are returning nil in all error cases and returning the same job in success. Is this return parameter really needed?

vsundarraj-px (approved):
looks good to me.

lalat-das:
closing the PR since the same PR with some more modification is raised against 2.7.2 now

PR #378, opened 2024-05-21
Enable kdmp & NFS  jobs to run in restricted PSA mode

lalat-das:
**What this PR does / why we need it**: This PR brings changes of psa and azure chine changes to kdmp via vending from stork to kdmp

Closes # pb-6681

Unit tested backup and restore with 1.2.13 executor image with corresponding stork PR

image

Restore of the above non-psi and psi backups are as below
image

### DIFF
--- a/pkg/drivers/utils/utils.go
+++ b/pkg/drivers/utils/utils.go
@@ -889,6 +889,29 @@ func IsJobPodMountFailed(job *batchv1.Job, namespace string) bool {
 	return false
 }
 
+// Check if a job has failed because of podSecurity violation
+func IsJobPodSecurityFailed(job *batchv1.Job, namespace string) bool {
+	fn := "IsJobPodSecurityFailed"
+
+	opts := metav1.ListOptions{
+		FieldSelector: "involvedObject.name=" + string(job.Name),
+	}
+	events, err := core.Instance().ListEvents(namespace, opts)
+	if err != nil {
+		errMsg := fmt.Sprintf("failed to fetch events for job [%s/%s]: %v", namespace, job.Name, err)
+		logrus.Debugf("%s: %v", fn, errMsg)
+		return false
+	}
+	// if the job event reason is Failedcreate due to fobidden podSecurity violation
+	// then return true
+	for _, event := range events.Items {
+		if event.Reason == "FailedCreate" && strings.Contains(event.Message, "violates PodSecurity") {
+			return true
+		}
+	}
+	return false
+}
+
 // DisplayJobpodLogandEvents - Prints the Job pod description, log and events
 func DisplayJobpodLogandEvents(jobName string, namespace string) {
 	// Get job from the namespace
@@ -976,7 +999,7 @@ func GetShortUID(uid string) string {
 }
 
 // Add container security Context to job pod if the PSA is enabled.
-// if static uids like kdmpJobUid or kdmpJobGid is used that means
+// If static uids like kdmpJobUid or kdmpJobGid is used that means
 // these are dummy UIDs used for backing up resources to backuplocation
 // which doesn't need specific UID specific permission.
 func AddSecurityContextToJob(job *batchv1.Job, podUserId, podGroupId string) (*batchv1.Job, error) {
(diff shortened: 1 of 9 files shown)


pkg/drivers/utils/utils.go
```diff
@@ -966,3 +997,91 @@ func GetShortUID(uid string) string {
...
+			return nil, fmt.Errorf("failed to convert the UID to int: %v", err)
+		}
+		job.Spec.Template.Spec.Containers[0].SecurityContext.RunAsUser = &uid
+
+		// Add fsgroup in Pod security context with the same UID as RunAsUser
+		// But we shouldn't add fsgroup if it is a kopia backup because it will alter the permission
+		// of the backup pod filesystem.
+		if !strings.Contains(job.Spec.Template.Spec.Containers[0].Command[0], kopiaBackupString) {
+			job.Spec.Template.Spec.SecurityContext = &corev1.PodSecurityContext{
+				FSGroup: &uid,
```
prashanthpx, line 1038: If it is no OCP backup where runAsUser, runAsGroup and FsGroup was set and restoring to another vanila cluster, should we not add the same to kdmp job pods? If we don't add wouldn't the job pod runs as root?

pkg/drivers/utils/utils.go
```diff
@@ -966,3 +997,91 @@ func GetShortUID(uid string) string {
...
+	if podGroupId != "" {
+		gid, err := strconv.ParseInt(podGroupId, 10, 64)
+		if err != nil {
+			logrus.Errorf("failed to convert the GID to int: %v", err)
+			return nil, fmt.Errorf("failed to convert the GID to int: %v", err)
+		}
+		job.Spec.Template.Spec.Containers[0].SecurityContext.RunAsGroup = &gid
+	}
+	// Add RunAsNonRoot to true and drop all capabilities and seccomp profile and allowPrivilegeEscalation to false
+	job.Spec.Template.Spec.Containers[0].SecurityContext.RunAsNonRoot = ptr.To(true)
```
prashanthpx, line 1051: For a normal bakcup where no security context was set, kdmp pod will run as root. With this check on the restore side, are making it run as non-root, what would be the user after restore? During backup they would be as `root`

prashanthpx (approved):
LGTM after addressing review comments

pkg/drivers/utils/utils.go
```diff
@@ -966,3 +997,91 @@ func GetShortUID(uid string) string {
 	}
 	return uid[:8]
 }
+
+// Add container security Context to job pod if the PSA is enabled.
+// if static uids like kdmpJobUid or kdmpJobGid is used that means
```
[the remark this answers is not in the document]
lalat-das: done.

pkg/drivers/utils/utils.go
lalat-das: So you are saying a scenario where in backup is taken for an app where in  security context is not there So job will run as **_root_** . During restore we will check if volume info In backup if there is a UID mentioned there won't be anything used in data export CR as annotation and we will not add any security context bloc.

pkg/drivers/utils/utils.go
lalat-das: In case of a non-ocp based backup and restore user group is preserved anyway and it will be used.. May be I need to understand the comment better. Let me discus this in Monday.

pkg/drivers/utils/utils.go
```diff
@@ -881,6 +889,29 @@ func IsJobPodMountFailed(job *batchv1.Job, namespace string) bool {
...
+	events, err := core.Instance().ListEvents(namespace, opts)
+	if err != nil {
+		errMsg := fmt.Sprintf("failed to fetch events for job [%s/%s]: %v", namespace, job.Name, err)
+		logrus.Debugf("%s: %v", fn, errMsg)
+		return false
+	}
+	// if the job event reason is Failedcreate due to fobidden podSecurity violation
+	// then return true
+	for _, event := range events.Items {
+		if event.Reason == "FailedCreate" && strings.Contains(event.Message, "violates PodSecurity") {
```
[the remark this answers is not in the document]
lalat-das: Yes they are same. Thanks for asking I tried in 1.27 k8s at least in that version it is same.

PR #388, opened 2024-07-18
PB-7476: Adding node affinity to kopia backup and restore jobs

shkumari-px:
**What this PR does / why we need it**: It adds node affinity rule to kopia backup and restore jobs

Closes # https://purestorage.atlassian.net/browse/PB-7476

The user needs to add this label to kdmp-config **PXB_JOB_NODE_AFFINITY_LABEL:kopia-backup**
This label has to be added to the application node pool as well
**Unit Testing**

1.  Add label to node and kdmp config **Node-2**
Screenshot 2024-08-01 at 10 27 02 AM

2.  Assigned label to a different node, other than node on which the mysql-pod is assigned which we be taking the backup of (mysql app with pvc running on **Node 3**)
mysqlappwithpvcNode

3. Took a backup of default namespace containing mysql app with pvc and noapp-pvc
Job pod created for mysql app with pvc got assigned to Node 3 with no node affinity in it's spec
livebackuppodwithoutnodeaffinity

**Updated result**
Screenshot 2024-08-06 at 1 12 08 PM

Job pod created for noapp-pvc got assigned to node 2 with node affinity in it's spec
nonlivebackupwithnodeafffinity

**Updated result**
Screenshot 2024-08-06 at 1 09 12 PM

**NFS backup Job Node Affinity**
Screenshot 2024-08-01 at 11 06 52 AM
**Updated Result**
Screenshot 2024-08-06 at 1 28 07 PM

**NFS Restore Pod**
Screenshot 2024-07-19 at 11 16 55 AM
**Updated Result**
Screenshot 2024-08-06 at 1 32 30 PM

**Kopia Restore Pod**
Screenshot 2024-07-19 at 11 19 30 AM

**Updated result**
noapp-pvc restore
Screenshot 2024-08-06 at 1 17 32 PM

mysql-pvc restore
Screenshot 2024-08-06 at 1 17 44 PM

csi backup on nfs bl
Screenshot 2024-08-05 at 5 57 13 PM
Screenshot 2024-08-05 at 6 01 49 PM

**nfscsirestore**
Screenshot 2024-08-05 at 6 06 07 PM
Screenshot 2024-08-05 at 6 09 05 PM
**Updated result**
Screenshot 2024-08-06 at 1 51 03 PM

Screenshot 2024-08-06 at 1 51 31 PM
Screenshot 2024-08-06 at 1 52 00 PM

Verified taking backup in psa **environment**
psa environmnet + postgress app with pvc  (backup successful)
psa environment - only pvc (backup failed)
Screenshot 2024-08-06 at 5 59 27 PM
Screenshot 2024-08-06 at 6 02 54 PM
**central ns psa environment**
Screenshot 2024-08-06 at 6 03 51 PM

### DIFF
--- a/pkg/controllers/dataexport/reconcile.go
+++ b/pkg/controllers/dataexport/reconcile.go
@@ -1903,6 +1903,11 @@ func startTransferJob(
 		psaJobUid = getAnnotationValue(dataExport, utils.PsaUIDKey)
 		psaJobGid = getAnnotationValue(dataExport, utils.PsaGIDKey)
 	}
+	nodeLabel, err := utils.GetNodeLabelFromDeployment(jobConfigMap, jobConfigMapNs, drivers.PxbJobNodeLabelKey)
+	if err != nil {
+		return "", err
+	}
+
 	switch drv.Name() {
 	case drivers.Rsync:
 		return drv.StartJob(
@@ -1947,6 +1952,7 @@ func startTransferJob(
 			drivers.WithExcludeFileList(excludeFileList),
 			drivers.WithPodDatapathType(podDataPath),
 			drivers.WithJobConfigMap(jobConfigMap),
+			drivers.WithNodeAffinity(nodeLabel),
 			drivers.WithJobConfigMapNs(jobConfigMapNs),
 			drivers.WithNfsServer(nfsServerAddr),
 			drivers.WithNfsExportDir(nfsExportPath),
@@ -1969,6 +1975,7 @@ func startTransferJob(
 			drivers.WithCertSecretNamespace(dataExport.Spec.Destination.Namespace),
 			drivers.WithJobConfigMap(jobConfigMap),
 			drivers.WithJobConfigMapNs(jobConfigMapNs),
+			drivers.WithNodeAffinity(nodeLabel),
 			drivers.WithNfsServer(nfsServerAddr),
 			drivers.WithNfsExportDir(nfsExportPath),
 			drivers.WithPodUserId(psaJobUid),
@@ -2396,6 +2403,11 @@ func startNfsCSIRestoreVolumeJob(
 		logrus.Errorf("failed to create NFS cred secret: %v", err)
 		return "", fmt.Errorf("failed to create NFS cred secret: %v", err)
 	}
+	nodeLabel, err := utils.GetNodeLabelFromDeployment(jobConfigMap, jobConfigMapNs, drivers.PxbJobNodeLabelKey)
+	if err != nil {
+		return "", err
+	}
+
 	switch drv.Name() {
 	case drivers.NFSCSIRestore:
 		return drv.StartJob(
@@ -2410,6 +2422,7 @@ func startNfsCSIRestoreVolumeJob(
 			drivers.WithNfsSubPath(bl.Location.Path),
 			drivers.WithPodUserId(psaJobUid),
 			drivers.WithPodGroupId(psaJobGid),
+			drivers.WithNodeAffinity(nodeLabel),
 		)
 	}
 	return "", fmt.Errorf("unknown driver for nfs csi volume restore: %s", drv.Name())
--- a/pkg/controllers/resourceexport/reconcile.go
+++ b/pkg/controllers/resourceexport/reconcile.go
@@ -412,6 +412,12 @@ func startNfsResourceJob(
 		logrus.Errorf("failed to create NFS cred secret: %v", err)
 		return "", fmt.Errorf("failed to create NFS cred secret: %v", err)
 	}
+
+	nodeLabel, err := utils.GetNodeLabelFromDeployment(jobConfigMap, jobConfigMapNs, drivers.PxbJobNodeLabelKey)
+	if err != nil {
+		return "", err
+	}
+
 	switch drv.Name() {
 	case drivers.NFSBackup:
 		return drv.StartJob(
@@ -427,6 +433,7 @@ func startNfsResourceJob(
 			drivers.WithAppCRNamespace(re.Spec.Source.Namespace),
 			drivers.WithNamespace(re.Namespace),
 			drivers.WithResoureBackupName(re.Name),
+			drivers.WithNodeAffinity(nodeLabel),
 			drivers.WithResoureBackupNamespace(re.Namespace),
 			drivers.WithNfsMountOption(bl.Location.NFSConfig.MountOptions),
 			drivers.WithNfsExportDir(bl.Location.NFSConfig.SubPath),
@@ -445,6 +452,7 @@ func startNfsResourceJob(
 			drivers.WithAppCRNamespace(re.Spec.Source.Namespace),
 			drivers.WithNamespace(re.Namespace),
 			drivers.WithResoureBackupName(re.Name),
+			drivers.WithNodeAffinity(nodeLabel),
 			drivers.WithResoureBackupNamespace(re.Namespace),
 			drivers.WithNfsMountOption(bl.Location.NFSConfig.MountOptions),
 			drivers.WithNfsExportDir(bl.Location.NFSConfig.SubPath),
--- a/pkg/drivers/kopiabackup/kopiabackup.go
+++ b/pkg/drivers/kopiabackup/kopiabackup.go
@@ -273,6 +273,7 @@ func jobFor(
 	jobName string,
 	resources corev1.ResourceRequirements,
 	nodeName string,
+	live bool,
 ) (*batchv1.Job, error) {
 	backupName := jobName
 
@@ -410,6 +411,14 @@ func jobFor(
 		job.Spec.Template.Spec.ImagePullSecrets = utils.ToImagePullSecret(utils.GetImageSecretName(jobName))
 	}
 
+	// Add node affinity to the job spec
+	if !live {
+		job, err = utils.AddNodeAffinityToJob(job, jobOption)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	if len(jobOption.NfsServer) != 0 {
 		volumeMount := corev1.VolumeMount{
 			Name:      utils.NfsVolumeName,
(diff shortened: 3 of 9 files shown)


pkg/controllers/dataexport/reconcile.go
```diff
@@ -1869,11 +1869,12 @@ func startTransferJob(
...
-		nfsExportPath  string
-		nfsMountOption string
-		psaJobUid      string
-		psaJobGid      string
+		nfsServerAddr              string
+		nfsExportPath              string
+		nfsMountOption             string
+		psaJobUid                  string
+		psaJobGid                  string
+		kopiaBackupJobNodeLabelKey = "KOPIA_BACKUP_NODE_AFFINITY_LABEL"
```
px-kesavan, line 1877: can we move this var to pkg/drivers/drivers.go ?

pkg/controllers/dataexport/reconcile.go
```diff
@@ -1885,6 +1886,15 @@ func startTransferJob(
...
 	}
+	nodeLabel := make(map[string]string)
+	kdmpData, err := core.Instance().GetConfigMap(jobConfigMap, jobConfigMapNs)
+	if err != nil {
+		return "", err
+	}
+	kopiaBackupJobNodeLabelValue, ok := kdmpData.Data[kopiaBackupJobNodeLabelKey]
+	if ok && kopiaBackupJobNodeLabelValue != "" {
+		nodeLabel[kopiaBackupJobNodeLabelKey] = kopiaBackupJobNodeLabelValue
+	}
```
px-kesavan, line 1897: Can we have a generalized naming ?, as this nodelabel is used for restore as well

pkg/controllers/resourceexport/reconcile.go
```diff
@@ -405,6 +405,16 @@ func startNfsResourceJob(
...
+	var kopiaBackupJobNodeLabelKey = "KOPIA_BACKUP_NODE_AFFINITY_LABEL"
+	nodeLabel := make(map[string]string)
+	kdmpData, err := core.Instance().GetConfigMap(jobConfigMap, jobConfigMapNs)
+	if err != nil {
+		return "", err
+	}
+	kopiaBackupJobNodeLabelValue, ok := kdmpData.Data[kopiaBackupJobNodeLabelKey]
+	if ok && kopiaBackupJobNodeLabelValue != "" {
+		nodeLabel[kopiaBackupJobNodeLabelKey] = kopiaBackupJobNodeLabelValue
+	}
```
px-kesavan, line 417: Same here . Can we have a generalized naming ?, as this nodelabel is used for restore as well

pkg/drivers/kopiarestore/kopiarestore.go
```diff
@@ -306,6 +306,29 @@ func jobFor(
...
+	if len(jobOption.NodeAffinity) > 0 {
+		matchExpressions := []corev1.NodeSelectorRequirement{}
+		for key, val := range jobOption.NodeAffinity {
+			expression := corev1.NodeSelectorRequirement{
+				Key:      key,
+				Operator: corev1.NodeSelectorOpIn,
+				Values:   []string{val},
+			}
+			matchExpressions = append(matchExpressions, expression)
+		}
```
px-kesavan, line 318: same here

pkg/drivers/kopiabackup/kopiabackup.go
```diff
@@ -410,6 +410,31 @@ func jobFor(
...
+					NodeSelectorTerms: []corev1.NodeSelectorTerm{
+						{
+							MatchExpressions: matchExpressions,
+						},
+					},
+				},
+			},
+		}
+	}
+
```
px-kesavan, line 437: @shkumari-px One thing we need test on how this is impacted in RWO(readWriteOnce) PVs. When the pv of RWO access mode is attached to a node (say node 1) and with nodeaffinity we are trying to run the kdmp backup/restore jobs in another node (lets say node 2).

shkumari-px: > @shkumari-px One thing we need test on how this is impacted in RWO(readWriteOnce) PVs. When the pv of RWO access mode is attached to a node (say node 1) and with nodeaffinity we are trying to run the kdmp backup/restore jobs in another node (lets say node 2).

we get the below error. in this case 
Reason:FailedMount,Message:MountVolume.SetUp failed for volume "pvc-52586bcd-fe1f-4df1-aa70-ba831f096c4c" : rpc error: code = Internal desc = failed to attach volume: Non-shared volume is already attached on another node. Non-shared volumes can only be attached on one node at a time.,Source:EventSource{Component:kubelet,Host:shkumari-116-2,}

here the expectation is to fail the backup
We are adding the node affinity because one of the customer had 2 node pools

- ingress worker node pool (had network restriction)
- application  node pool (application run on this pool only)
here the kdmp jobs got scheduled on ingress worker node pool and the backups were failing

so to fix this we will be adding a label to application node pool and we will be using the same label to schedule the kdmp jobs, this is for non live kdmp jobs

prashanthpx: @shkumari-px To understand, when kdmp job was tyring to figure out which node was using the app, did it point to ingress worker sometimes and sometime app node pool? Can you add some details here?

siva-portworx: Actually she  has mentioned the input from the Ranjit in the ticket: https://purestorage.atlassian.net/browse/PB-3647?focusedCommentId=340726
Basically, for non-live PVC backups, customer want the job pods to be schedule on application node pool instead of ingress worker node pool. Since currently we do not have affinity associated to job pods, it lands  on any nodes.
@shkumari-px , @px-kesavan comment is to verify that non-live RWO PVC backup works fine or not.

@px-kesavan , Answer to your question is that PVC in bounded state does not mean that it is attach to any node.  It is just an association to a volume. Only when some pods tries to mount it, it will be get attached to a node and mounted.

```
A PVC (Persistent Volume Claim) being bound means that it has been successfully associated with a Persistent Volume (PV). This does not necessarily mean the volume is attached to a node or mounted in a pod.

Here are the stages involved:

PVC Bound to PV: When a PVC is bound, it means that the Kubernetes control plane has found a suitable PV that matches the PVC’s requirements and has associated them. At this stage, the volume is allocated but not yet attached to any node or mounted by any pod.

Volume Attached to Node: When a pod that references the PVC is scheduled onto a node, the volume is then attached to that node. This means the underlying storage is connected to the node, making it accessible to the node's filesystem.

Volume Mounted in Pod: The final step is when the volume is mounted inside the pod. This happens when the pod is started, and the container(s) within the pod have the specified volume mounted at the designated path within the container's filesystem.

To summarize:

Bound: PVC is matched to a PV.
Attached: Volume is connected to a node.
Mounted: Volume is accessible inside the pod's container.


```

pkg/drivers/kopiabackup/kopiabackup.go
```diff
@@ -519,10 +546,11 @@ func buildJob(jobName string, jobOptions drivers.JobOpts) (*batchv1.Job, error)
 		jobName,
 		resources,
 		nodeName,
+		live,
 	)
 }
 
-func roleFor(live bool) *rbacv1.Role {
+func roleFor() *rbacv1.Role {
```
vsundarraj-px, line 553: why is this removed? Not able to understand the relevance of this change for this fix.

pkg/drivers/kopiabackup/kopiabackup.go
```diff
@@ -532,22 +560,5 @@ func roleFor(live bool) *rbacv1.Role {
 			},
 		},
 	}
-	// Only live backup, we will add the hostaccess and privilege option.
```
shkumari-px, line 535: @vsundarraj-px This code was never used
We had a live bool flag but it was never set to true in the previous already existing code.
Since now we will be setting the live flag to true, this if condition will get triggered now which we don't want
Also this piece of code was unnecessary

vsundarraj-px: This seems PSA code to me can you confirm this with @lalat-das. @shkumari-px I would prefer a different variable to use here instead of same variable and changing some dead code to avoid confusion. 
If this is a dead code, I think you should remove it as part of another commit not as part of same commit.

siva-portworx: Or else add a commit message about the cleanup in the same commit as well.

siva-portworx:
Should we also do node affinity for nfscsirestore job as well?

pkg/drivers/kopiabackup/kopiabackup.go
```diff
@@ -410,6 +410,31 @@ func jobFor(
...
+	if len(jobOption.NodeAffinity) > 0 {
+		matchExpressions := []corev1.NodeSelectorRequirement{}
+		for key, val := range jobOption.NodeAffinity {
+			expression := corev1.NodeSelectorRequirement{
+				Key:      key,
+				Operator: corev1.NodeSelectorOpIn,
+				Values:   []string{val},
+			}
+			matchExpressions = append(matchExpressions, expression)
+		}
```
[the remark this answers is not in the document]
shkumari-px: we can support multiple key-value pairs of node label

pkg/drivers/kopiabackup/kopiabackup.go
shkumari-px: I'll add a separate commit for this cleanup

prashanthpx (approved):
LGTM

kgarg-px (approved):

pkg/drivers/utils/utils.go
```diff
@@ -855,6 +855,20 @@ func GetNodeAffinityFromDeployment(name, namespace string) (*corev1.NodeAffinity
 	return deploy.Spec.Template.Spec.Affinity.NodeAffinity, nil
 }
 
+// Get node lablel from deployment
```
aks-px, line 858: Typo -> label

aks-px:
```suggestion
// GetNodeLabelFromDeployment gets node label from deployment
```

aks-px (approved):
LGTM

shkumari-px:
> Should we also do node affinity for nfscsirestore job as well?

done
Took a backup of thin-csi pvc with thin-csi-sc on nfs backup location
Took a restore of this backup and validated that the node affinity is populated in the restore pod spec
Have attached the screenshot

siva-portworx (approved):

PR #389, opened 2024-07-18
pb-7598: Removed default addition of the RunAsGroup in the securityAc…

siva-portworx:
```
pb-7598: Removed default addition of the RunAsGroup in the securityAccount.
```
Closes # pb-7598

Testing:
Tested the backup/restore with OCP setup of application that had RDB volume with offload option, which was failing previous with permission denied error.

Screenshot 2024-07-18 at 3 31 42 PM
Screenshot 2024-07-18 at 3 31 33 PM

### DIFF
--- a/pkg/drivers/nfsbackup/nfsbackup.go
+++ b/pkg/drivers/nfsbackup/nfsbackup.go
@@ -281,7 +281,9 @@ func jobForBackupResource(
 	// The Job is intended to backup resources to  NFS backuplocation
 	// and it doesn't need a specific JOB uid/gid since it will be sqaushed at NFS server
 	// hence used a global hardcoded UID/GID.
-	job, err = utils.AddSecurityContextToJob(job, utils.KdmpJobUid, utils.KdmpJobGid)
+	// Not passing the groupId as we do not want to set the RunAsGroup field in the securityContext
+	// This helps us in setting the primaryGroup ID to root for the user ID.
+	job, err = utils.AddSecurityContextToJob(job, utils.KdmpJobUid, "")
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/drivers/nfsrestore/nfsrestore.go
+++ b/pkg/drivers/nfsrestore/nfsrestore.go
@@ -321,7 +321,9 @@ func jobForRestoreResource(
 			},
 		},
 	}
-	job, err = utils.AddSecurityContextToJob(job, utils.KdmpJobUid, utils.KdmpJobGid)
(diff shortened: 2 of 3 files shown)


t-nidhin (approved):

pkg/drivers/nfsbackup/nfsbackup.go
```diff
@@ -281,7 +281,9 @@ func jobForBackupResource(
 	// The Job is intended to backup resources to  NFS backuplocation
 	// and it doesn't need a specific JOB uid/gid since it will be sqaushed at NFS server
 	// hence used a global hardcoded UID/GID.
-	job, err = utils.AddSecurityContextToJob(job, utils.KdmpJobUid, utils.KdmpJobGid)
+	// Not passing the groupId as we do not want to set the RunAsGroup field in the securityContext
+	// This helps us in setting the primaryGroup ID to root for the user ID.
+	job, err = utils.AddSecurityContextToJob(job, utils.KdmpJobUid, "")
```
vsundarraj-px, line 286: Isnt this countering the purpose of setting the runAsGroup although dummy for non OCP case too. I see that that fix is intended to not set runAsGroup in case of NFS on OCP, but passing "" for all case will cause regression?

prashanthpx (approved):

pkg/drivers/nfsbackup/nfsbackup.go
siva-portworx: Not setting RunAsGroups, such that we will make the user to be part of root group and it is helping us, if some of the provisioner do not respect the FsGroup and make the owner group as root instead of Fsgroup id.
This is working in PSA environment as well. So we are doing this. This is how the default OCP securityContext is set currently.

pkg/drivers/nfsbackup/nfsbackup.go
```diff
@@ -281,7 +281,9 @@ func jobForBackupResource(
 	// The Job is intended to backup resources to  NFS backuplocation
 	// and it doesn't need a specific JOB uid/gid since it will be sqaushed at NFS server
 	// hence used a global hardcoded UID/GID.
-	job, err = utils.AddSecurityContextToJob(job, utils.KdmpJobUid, utils.KdmpJobGid)
+	// Not passing the groupId as we do not want to set the RunAsGroup field in the securityContext
+	// This helps us in setting the primaryGroup ID to root for the user ID.
```
vsundarraj-px, line 285: nit: Earlier comment and new comment are missleading. We may have to ensure one of them is valid.

PR #390, opened 2024-07-24
PB-7645: Add pause to cleanup of resources if env is set in kdmp-config configmap

aks-px:
…

This PR pauses the cleanup that happens after the job is completed for the specified time in the PAUSE_RESOURCE_CLEANUP key.This is helpful incase of debugging to persist the resources for time mentioned 

Closes #PB-7645

Tested scenarios with value and without the key in the config map,working as expected:

- Without the key,cleanup is happening soon after it goes to completed state
Screenshot 2024-07-24 at 11 44 15 AM
Screenshot 2024-07-24 at 11 43 17 AM

- After adding key in configmap:
Screenshot 2024-07-24 at 11 35 07 AM
Screenshot 2024-07-24 at 11 18 12 AM
Screenshot 2024-07-24 at 11 22 25 AM

After the review comments:
Screenshot 2024-07-26 at 5 57 40 PM (3)

### DIFF
--- a/pkg/controllers/dataexport/reconcile.go
+++ b/pkg/controllers/dataexport/reconcile.go
@@ -1131,7 +1131,13 @@ func (c *Controller) stageLocalSnapshotRestore(ctx context.Context, dataExport *
 		}
 		return true, c.updateStatus(dataExport, data)
 	} else if dataExport.Status.Status == kdmpapi.DataExportStatusFailed {
-		err := c.cleanupLocalRestoredSnapshotResources(dataExport, false)
+		val, err := utils.PauseCleanupResource()
+		if err == nil && val != 0 {
+			logrus.Debugf("Starting to wait for %v before cleanup of local restored snapshot resources for dataexport:%v", val, dataExport.Name)
+			time.Sleep(val)
+		}
+		logrus.Debugf("Starting the cleanup of local restored snapshot resources for dataexport:%v", dataExport.Name)
+		err = c.cleanupLocalRestoredSnapshotResources(dataExport, false)
 		if err != nil {
 			logrus.Errorf("cleaning up temporary resources for restoring from snapshot failed for data export %s/%s: %v", dataExport.Namespace, dataExport.Name, err)
 		}
@@ -1284,7 +1290,13 @@ func (c *Controller) stageLocalSnapshotRestoreInProgress(ctx context.Context, da
 		}
 		return true, c.updateStatus(dataExport, data)
 	} else if dataExport.Status.Status == kdmpapi.DataExportStatusFailed {
-		err := c.cleanupLocalRestoredSnapshotResources(dataExport, false)
+		val, err := utils.PauseCleanupResource()
+		if err == nil && val != 0 {
+			logrus.Debugf("Starting to wait for %v before cleanup of local restored snapshots for dataexport:%v", val, dataExport.Name)
+			time.Sleep(val)
+		}
+		logrus.Debugf("Starting the cleanup of local restored snapshot resources for dataexport:%v", dataExport.Name)
+		err = c.cleanupLocalRestoredSnapshotResources(dataExport, false)
 		// Already done with max retries, so moving to kdmp restore anyway
 		if err != nil {
 			logrus.Errorf("cleaning up temporary resources for restoring from snapshot failed for data export %s/%s: %v", dataExport.Namespace, dataExport.Name, err)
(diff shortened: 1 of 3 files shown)


pkg/drivers/utils/utils.go
```diff
@@ -1088,3 +1090,20 @@ func GetOcpNsUidGid(nsName string, psaJobUid string, psaJobGid string) (string,
 	}
 	return psaJobUid, psaJobGid, isOcp, nil
 }
+
+// PauseCleanupResource returns whether to pause the cleanup the CRs & other resources.
+func PauseCleanupResource() (time.Duration, error) {
+	pauseCleanupVal := time.Duration(0)
+	pausecleanupValStr, err := k8sutils.GetConfigValue(KdmpConfig, defaultPXNamespace, PauseResourceCleanupKey)
```
kgarg-px, line 1097:
```suggestion
	pauseCleanupValStr, err := k8sutils.GetConfigValue(KdmpConfig, defaultPXNamespace, PauseResourceCleanupKey)
```
Let's keep the naming consistent

pkg/drivers/utils/utils.go
```diff
@@ -1088,3 +1090,20 @@ func GetOcpNsUidGid(nsName string, psaJobUid string, psaJobGid string) (string,
...
+	if err != nil {
+		logrus.Errorf("Failed to get %s key from kdmp-config-map: %v", PauseResourceCleanupKey, err)
+		return pauseCleanupVal, err
+	}
+	if pausecleanupValStr != "" {
```
kgarg-px, line 1102: Same here. Let's keep the naming consistent

pkg/controllers/dataexport/reconcile.go
```diff
@@ -1131,7 +1131,16 @@ func (c *Controller) stageLocalSnapshotRestore(ctx context.Context, dataExport *
...
-		err := c.cleanupLocalRestoredSnapshotResources(dataExport, false)
+		val, err := utils.PauseCleanupResource()
+		if err != nil {
+			// since default value is 0,do nothing.
+		} else {
+			if val != 0 {
+				logrus.Infof("Starting to wait for %v minutes before cleanup of local restored snapshot resources:%v", val, dataExport.Name)
+				time.Sleep(val)
+			}
+		}
```
kgarg-px, line 1142:
```suggestion
		val, err := utils.PauseCleanupResource()
		if err == nil && val != 0 {
			logrus.Infof("Starting to wait for %v minutes before cleanup of local restored snapshot resources:%v", val, dataExport.Name)
			time.Sleep(val)
		}
```

Since the above `err!=nil` does not do anything, its better to keep the code more concise and readable

pkg/controllers/dataexport/reconcile.go
```diff
@@ -1284,7 +1293,16 @@ func (c *Controller) stageLocalSnapshotRestoreInProgress(ctx context.Context, da
...
```
kgarg-px, line 1304: Please avoid using nested loops if possible for better readability. Refer to the comment above

pkg/controllers/dataexport/reconcile.go
```diff
@@ -1455,6 +1473,15 @@ func (c *Controller) cleanUp(driver drivers.Interface, de *kdmpapi.DataExport) e
...
 	}
+	val, err := utils.PauseCleanupResource()
+	if err != nil {
+		// since default value is 0,do nothing.
+	} else {
+		if val != 0 {
+			logrus.Infof("Starting to wait for %v minutes before cleanup", val)
+			time.Sleep(val)
+		}
+	}
```
kgarg-px, line 1484: Same as rest

pkg/controllers/resourceexport/reconcile.go
```diff
@@ -245,6 +246,15 @@ func (c *Controller) process(ctx context.Context, in *kdmpapi.ResourceExport) (b
...
 func (c *Controller) cleanupResources(resourceExport *kdmpapi.ResourceExport) error {
+	val, err := utils.PauseCleanupResource()
+	if err != nil {
+		// since default value is 0,do nothing.
+	} else {
+		if val != 0 {
+			logrus.Infof("Starting to wait for %v minutes before cleanup of resourceExport:%v", val, resourceExport.Name)
+			time.Sleep(val)
+		}
+	}
```
kgarg-px, line 257: Same here as well

kgarg-px (commented):
Logic seems to be good, left a few nits

pkg/controllers/dataexport/reconcile.go
aks-px: Thanks for the suggesstion,will check on it

kgarg-px (approved):
LGTM!

pkg/controllers/dataexport/reconcile.go
```diff
@@ -1131,7 +1131,12 @@ func (c *Controller) stageLocalSnapshotRestore(ctx context.Context, dataExport *
 		}
 		return true, c.updateStatus(dataExport, data)
 	} else if dataExport.Status.Status == kdmpapi.DataExportStatusFailed {
-		err := c.cleanupLocalRestoredSnapshotResources(dataExport, false)
+		val, err := utils.PauseCleanupResource()
+		if err == nil && val != 0 {
+			logrus.Infof("Starting to wait for %v minutes before cleanup of local restored snapshot resources for dataexport:%v", val, dataExport.Name)
```
prashanthpx, line 1136: I prefer not to have this msg by default even when the value is zero (as this msg will be default printed always). Lets only have this when the value is really set.

prashanthpx (approved):
LGTM

PR #391, opened 2024-07-30
pb-7504: make NFS job pod to use root for resource backup

lalat-das:
- When we use GCP based file store as NFS backup location, the job pod using that doesn't have write permission for group user, this causes the non-root user permission denied error during backup and restore.

- This is GKE specific behaviour hence a check added to force all job pod to run as a root user eradicating the permission denied error.

**What this PR does / why we need it**: This PR fixes the GKE resource backup and restore path wherein if NFS backup location is used then it will not use non-root user in any case. Hence no security context will be added.

Closes # pb-7504

Tested with kdmp volume + resource and then resource only 

image

restore of the same 

image

### DIFF
--- a/pkg/drivers/utils/utils.go
+++ b/pkg/drivers/utils/utils.go
@@ -75,6 +75,8 @@ const (
 	OcpUidRangeAnnotationKey    = "openshift.io/sa.scc.uid-range"
 	OcpGidRangeAnnotationKey    = "openshift.io/sa.scc.supplemental-groups"
 	kopiaBackupString           = "kopiaexecutor backup"
+	// if providerType in node spec has this string then it is GCP hosted cluster
+	GCPBasedClusterString = "gce://"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/drivers/nfsrestore/nfsrestore.go
```diff
@@ -321,9 +321,25 @@ func jobForRestoreResource(
...
 	}
+
+	uid := utils.KdmpJobUid
+	// For GCP based clusters the NFS PVC mounted with a anomalous GID permissions( i.e. sans GID write permission)
+	// hence avoiding passing any specific UID or GID so that Job pod will always run as ROOT user.
+	// This makes the job pod to fail in GCP based cluster with PSA enabled environment.
+
+	// check the cluster is GCP based or not
+	isGcpBasedCluster, err := utils.IsGcpHostedCluster()
+	if err != nil {
```
siva-portworx, line 332: Do we need this fix for restore as well? for restore, we need only read permission rt?

pkg/drivers/utils/utils.go
```diff
@@ -1088,3 +1088,21 @@ func GetOcpNsUidGid(nsName string, psaJobUid string, psaJobGid string) (string,
...
+	// Any GCP hosted cluster be it vanilla , OCP or GKE
+	// it is expected to have a ProviderId in its spec witha prefix of "gce"
+	nodes, err := core.Instance().GetNodes()
+	if err != nil {
+		return false, fmt.Errorf("failed to get nodes: %v", err)
+	}
+
+	for _, node := range nodes.Items {
+		providerID := node.Spec.ProviderID
+		if strings.HasPrefix(providerID, "gce://") {
```
siva-portworx, line 1103: nit: can we make the "gce://" as string constant definition?

pkg/drivers/nfsrestore/nfsrestore.go
lalat-das: good catch we don't need it.

pkg/drivers/utils/utils.go
```diff
@@ -1088,3 +1088,21 @@ func GetOcpNsUidGid(nsName string, psaJobUid string, psaJobGid string) (string,
 	}
 	return psaJobUid, psaJobGid, isOcp, nil
 }
+
+// Checks if the cluster is GCP hosted cluster.
+func IsGcpHostedCluster() (bool, error) {
+	// Any GCP hosted cluster be it vanilla , OCP or GKE
+	// it is expected to have a ProviderId in its spec witha prefix of "gce"
```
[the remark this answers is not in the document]
lalat-das: done

pkg/drivers/utils/utils.go
lalat-das: done.

siva-portworx (approved):

PR #397, opened 2024-09-19
PB-8293 : Implement a new log writer which flushes the logs in every 30 seconds (for real time logs)

px-kesavan:
…30 seconds (for real time logs)

**What this PR does / why we need it**: Currently the output of kopia backup/restore command is flushed in to the pod logs only after the cmd execution , so the user is not aware of the progress This PR Implements a new log writer which flushes the logs in every 30 seconds (for real time logs)

Closes # https://purestorage.atlassian.net/browse/PB-8293

Output for backup 

Screenshot from 2024-09-19 10-28-56

Attached additional testing details with logs in the ticket https://purestorage.atlassian.net/browse/PB-8293

### DIFF
--- a/pkg/kopia/backup.go
+++ b/pkg/kopia/backup.go
@@ -4,12 +4,19 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
+	"log"
 	"os"
 	"os/exec"
+	"sync"
 	"time"
 
 	cmdexec "github.com/portworx/kdmp/pkg/executor"
 	"github.com/sirupsen/logrus"
+)
+
+const (
+	commandExecLogInterval = 30 * time.Second
 )
 
 // BackupSummaryResponse describes single snapshot entry.
@@ -74,6 +81,34 @@ type backupExecutor struct {
 	lastError       error
 }
 
+type logWriter struct {
+	logger      *log.Logger
+	lastLogTime time.Time
+	interval    time.Duration
+	mu          sync.Mutex
+}
+
+func newLogWriter(l *log.Logger, interval time.Duration) *logWriter {
+	return &logWriter{
+		logger:   l,
+		interval: interval,
+	}
+}
+
+func (lw *logWriter) Write(p []byte) (n int, err error) {
+	lw.mu.Lock()
+	defer lw.mu.Unlock()
+
+	now := time.Now()
+	// log if the interval has passed since the last log
+	// this is to avoid bloating the logs
+	if now.Sub(lw.lastLogTime) >= lw.interval {
+		lw.logger.Println(string(p), time.Now())
+		lw.lastLogTime = now
+	}
+	return len(p), nil
+}
+
 // GetBackupCommand returns a wrapper over the kopia backup command
 func GetBackupCommand(path, repoName, password, provider, sourcePath string) (*Command, error) {
 	if repoName == "" {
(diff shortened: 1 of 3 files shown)


pkg/kopia/backup.go
```diff
@@ -74,6 +81,34 @@ type backupExecutor struct {
...
+
+func newLogWriter(l *log.Logger, interval time.Duration) *logWriter {
+	return &logWriter{
+		logger:   l,
+		interval: interval,
+	}
+}
+
+func (lw *logWriter) Write(p []byte) (n int, err error) {
+	lw.mu.Lock()
```
lalat-das, line 99: Why we need this mutex. What is the scenario when write will be called from two parallel path.
Isn't multi writer already takes care of writing to two  file descriptor  in parallel with synchronization on.

pkg/kopia/backup.go
```diff
@@ -74,6 +81,34 @@ type backupExecutor struct {
...
+	defer lw.mu.Unlock()
+
+	now := time.Now()
+	// log if the interval has passed since the last log
+	// this is to avoid bloating the logs
+	if now.Sub(lw.lastLogTime) >= lw.interval {
```
lalat-das, line 105: What will be the behavior if backup gets completed with 30 sec. Nothing will be written ? Say some error in the command line happens and command is done. I am confused about the _`lastLogTime`_'s value when for the first time the write call is invoked ? Any default value need to be assigned?  or is it because it will be zero and So `now - 0` already takes care of it?

lalat-das (commented):
please address the review Comment.

pkg/kopia/backup.go
```diff
@@ -74,6 +81,34 @@ type backupExecutor struct {
...
+		lw.logger.Println(string(p), time.Now())
```
lalat-das, line 106: Please add a test case showcasing **_failure of kopia command_** if it can happen within 30 sec and log appears as expected then we are good.

pkg/kopia/backup.go
px-kesavan: Even though we create two separate objects by calling newLogWriter for stderr and stdout, there is a chance that write can be called by multiple goroutines at the same time if the snapshot create command generates output quickly. It's unlikely to happen, but I have added a mutex for extra integrity of the output data

pkg/kopia/backup.go
px-kesavan: First write will not wait for 30secs, it will be written as when it comes. 
To explain on the lastLogTime value for the first time will be 0 time format. 
if block basically does (time.Now() - lastLogTime ) > 30 sec, Which will succeed .

pkg/kopia/backup.go
px-kesavan: Please refer the ticket, have attached for the same

pkg/kopia/backup.go
```diff
@@ -103,8 +138,13 @@ func NewBackupExecutor(cmd *Command) Executor {
 
 func (b *backupExecutor) Run() error {
 	b.execCmd = b.cmd.BackupCmd()
-	b.execCmd.Stdout = b.outBuf
-	b.execCmd.Stderr = b.errBuf
+
+	// Create multi-writers to stream output to both buffer and CLI
+	stdoutWriter := io.MultiWriter(b.outBuf, newLogWriter(log.New(os.Stdout, "", 0), commandExecLogInterval))
+	stderrWriter := io.MultiWriter(b.errBuf, newLogWriter(log.New(os.Stderr, "", 0), commandExecLogInterval))
```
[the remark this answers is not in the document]
px-kesavan: Not doubting , but still doubting :) . Have attached the logs for the same in the ticket. Thanks

lalat-das (approved):
lgtm.

PR #399, opened 2024-10-18
PB-8347: Fix KDMP Backup Failure in-case KDMP job removed post VB CR is updated with SnapshotID

sgajawada-px:
	- Update CreateVolumeBackup to return a boolean based upon the snapshotID
	- Update KopiaBackup runBackup to exit successfully in case the VB CR already holds the snapshotID
	
Closes #PB-8347

PFA of Testing Screenshots
Screenshot from 2024-10-17 20-15-27
Screenshot from 2024-10-17 20-15-16
Screenshot from 2024-10-17 20-14-55
Screenshot from 2024-10-17 20-14-39

### DIFF
--- a/pkg/executor/common.go
+++ b/pkg/executor/common.go
@@ -607,7 +607,7 @@ func UpdateResourceBackupStatus(
 }
 
 // CreateVolumeBackup creates volumebackup CRD
-func CreateVolumeBackup(name, namespace, repository, blName, blNamespace string) error {
+func CreateVolumeBackup(name, namespace, repository, blName, blNamespace string) (bool, error) {
 	new := &kdmpapi.VolumeBackup{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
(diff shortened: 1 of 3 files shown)


pkg/executor/common.go
```diff
@@ -607,7 +607,7 @@ func UpdateResourceBackupStatus(
 }
 
 // CreateVolumeBackup creates volumebackup CRD
-func CreateVolumeBackup(name, namespace, repository, blName, blNamespace string) error {
+func CreateVolumeBackup(name, namespace, repository, blName, blNamespace string) (bool, error) {
```
vikasit12, line 610: Is it necessary to change function signature?

Can't simply return nil if snaphotID is already there.

`if volumeBackup.Status.SnapshotID != "" {
    // If SnapshotID exists, return early or handle accordingly
    return nil 
}
// Proceed with the rest of the logic if SnapshotID does not exist`

sgajawada-px: If we return nil, then will go with the control flow and takes another snapshot (redoing the same work again). So instead will return from the runBackup to prevent taking another snapshot.

vikasit12 (approved):

